PR #1, opened 2021-05-11
Add conda packaging

lgruen:
@vladsaveliev FYI


.bumpversion.cfg
```diff
@@ -0,0 +1,6 @@
+[bumpversion]
+current_version = 0.0.1
+commit = True
+tag = True
```
illusional: We've left `tag = False` in the past, because we don't want to release a tag, as often it's tagged against a branch which might not contain everything.

.github/workflows/condarise.yaml
```diff
@@ -0,0 +1,56 @@
+name: Condarise
+on:
+  push:
+    branches:
+      - main
```
illusional, line 5: In analysis-runner, we only release pushes to master that bump the version: https://github.com/populationgenomics/analysis-runner/blob/8cecf118da39a1153f5668cbbdef3096d97948be/.github/workflows/condarise.yaml#L9

Might be good to do that here, unless the version is suffix-ed by the tag or something.

setup.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python
+
+import setuptools
+
+setuptools.setup(
+    name='cpg-utils',
+    version='0.0.1',
```
illusional, line 8:
```suggestion
    # This tag is automatically updated by bump2version
    version='0.0.1',
```

README.md
```diff
@@ -1,2 +1,16 @@
 # cpg-utils
-Library of convenience functions specific to the CPG
+
+This is a Python library containing convenience functions that are specific to the CPG.
+
+On every merge with the `main` branch, a new version gets published in the `cpg` conda
```
illusional: If you make above changes re releasing on bump2version commits, then this line would also have to change.

illusional (changes requested):
The merge into master might cause issues if there's no suffix (as the version might not change). The structure looks pretty good, if we find we have so many functions in one file that it's becoming a burden, we could also make it a folder:

```
cloud/
    __init__.py     # contains imports or some functions
    auth.py         # get email from request, etc...
```

Then `setup.py` just changes to include `cloud` as a package.

.github/workflows/condarise.yaml
lgruen: Perfect, thanks! I've now just copied the analysis-runner `condarise` workflow.

.bumpversion.cfg
lgruen: Thanks, done!

README.md
lgruen: Thank you!

illusional (approved):
🎉

PR #7, opened 2021-08-03
Add sample-metadata foundations

illusional:
- Adds requests for getting sample-id-maps

Todo:
- Matrix table annotation / column renaming
- A better cache


illusional:
I'm going to close this because it's outdated, and it hasn't really been an issue yet.

PR #24, opened 2022-04-05
adds git functions from analysis-runner

### LINKED ISSUE #23: Migrate git.py from analysis-runner to cpg-utils
The git.py code from analysis-runner (https://github.com/populationgenomics/analysis-runner/blob/main/analysis_runner/git.py) contains helpful functionality for cloning and managing git repositories within hail batches. To cut the dependency on analysis-runner, migrate that functionality into this repository. 

Changes: 
- remove organisation lock to `populationgenomics`?
- remove the in-container failure for the commit being on main?
- remove lock to single key file location

MattWellie:
closes #23 

- removes specific organisation check
- retains `commit-on-main | is_test` check
- auth key location can be altered (default value retained)

### DIFF
--- a/cpg_utils/git.py
+++ b/cpg_utils/git.py
@@ -148,7 +148,7 @@ def prepare_git_job(
         * Check out the specific commit
 
     Requires the authentication command to be executed prior to calling this method
-        - hail.gcloud_authenticate_job
+        - hail.authenticate_job
 
     Parameters
     ----------
--- a/cpg_utils/git.py
+++ b/cpg_utils/git.py
@@ -169,7 +169,9 @@ def prepare_git_job(
     # Any job commands here are evaluated in a bash shell, so user arguments should
     # be escaped to avoid command injection.
     job.command(
-        quote(f'git clone --recurse-submodules https://github.com/{organisation}/{repo_name}.git')
+        quote(
+            f'git clone --recurse-submodules https://github.com/{organisation}/{repo_name}.git'
+        )
     )
     job.command(f'cd {quote(repo_name)}')
 
--- a/cpg_utils/git.py
+++ b/cpg_utils/git.py
@@ -97,15 +97,20 @@ def get_repo_name_from_remote(remote_name: str) -> str:
     """
 
     repo = None
-    if remote_name.startswith('http'):
-        match = re.match(r'https:\/\/[A-z0-9\.]+?\/(.+?)\/(.+)$', remote_name)
-        organization, repo = match.groups()
+    try:
+        if remote_name.startswith('http'):
+            match = re.match(r'https:\/\/[A-z0-9\.]+?\/.+?\/(?P<repo>.+)$', remote_name)
+            if match:
+                repo = match.group('repo')
 
-    elif remote_name.startswith('git@'):
-        match = re.match(r'git@[A-z0-9\.]+?:(.+?)\/(.+)$', remote_name)
-        organization, repo = match.groups()
+        elif remote_name.startswith('git@'):
+            match = re.match(r'git@[A-z0-9\.]+?:.+?\/(?P<repo>.+)$', remote_name)
+            if match:
+                repo = match.group('repo')
+    except AttributeError as ae:
+        raise Exception(f'Unsupported remote format: "{remote_name}"') from ae
 
-    if not repo:
+    if repo is None:
         raise Exception(f'Unsupported remote format: "{remote_name}"')
 
     if repo.endswith('.git'):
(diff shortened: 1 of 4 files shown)


cpg_utils/git.py
```diff
@@ -0,0 +1,190 @@
...
+    job.env('AUTH_KEY_PATH', key_path)
+
+    # Use "set -x" to print the commands for easier debugging.
+    if print_all_statements:
+        job.command('set -x')
+
+    # activate the google service account
+    job.command(
+        f'gcloud -q auth activate-service-account --key-file=$AUTH_KEY_PATH'
+    )
```
lgruen: Might be cleaner to move this out of this function, as it's not really related to the `git` logic?

cpg_utils/git.py
```diff
@@ -0,0 +1,190 @@
...
+        return False
+
+
+def prepare_git_job(
+    job,
+    organisation: str,
+    repo_name: str,
+    commit: str,
+    is_test: bool = True,
+    print_all_statements: bool =True,
```
lgruen:
```suggestion
    print_all_statements: bool = True,
```

lgruen (changes requested):

lgruen:
You'll probably also want to add a minor version bump commit, so we can get this released in a new library.

MattWellie:
Try/Except surround, and `if match:` to handle pylint:
`Item "None" of "Optional[Match[str]]" has no attribute "group"`

cpg_utils/git.py
```diff
@@ -0,0 +1,188 @@
...
+):
+    """
+    Takes a hail batch job, and:
+        * Clones the repository
+            * if access_level != "test": check the desired commit is on 'main'
+        * Check out the specific commit
+
+    Requires the authentication command to be executed prior to calling this method
+        - hail.gcloud_authenticate_job
+
```
lgruen:
```suggestion
```

cpg_utils/hail.py
```diff
@@ -187,3 +187,36 @@ def image_path(suffix: str) -> str:
     prefix = os.getenv('CPG_IMAGE_REGISTRY_PREFIX')
     assert prefix
     return f'{prefix}/{suffix}'
+
+
+def gcloud_authenticate_job(
```
lgruen: In preparation for making this reusable across clouds, can we remove the `gcloud` here? Otherwise we'll have to change all call-sites later once we add Azure support.

lgruen (changes requested):

cpg_utils/hail.py
MattWellie: So reduce this to 'authenticate_job', and in future the same named method can work out how to do the auth (e.g. env variable?)

lgruen: Yes, exactly. We'll also have the added complication of having to authenticate gcloud even from Azure (to talk to the sample metadata server), but I hope that won't change the function declaration (right, @illusional?). We probably don't need to be able to override `key_path: str = '/gsa-key/key.json'` either, as that's really built into Hail Batch.

cpg_utils/hail.py
```diff
@@ -187,3 +187,36 @@ def image_path(suffix: str) -> str:
...
+    print_all_statements
+        * logging toggle
+
+    Returns
+    -------
+    None
+    """
+
+    # set the key path variable inside the job environment
+    job.env('AUTH_KEY_PATH', '/gsa-key/key.json')
```
lgruen: I don't have a strong opinion here, but could also just inline that in the `command`.

lgruen (approved):
Thanks for making this change, this looks good to me! Would be great to get @illusional's review as well though, as he knows this code a little better.

cpg_utils/hail.py
```diff
@@ -187,3 +187,33 @@ def image_path(suffix: str) -> str:
     prefix = os.getenv('CPG_IMAGE_REGISTRY_PREFIX')
     assert prefix
     return f'{prefix}/{suffix}'
+
+
+def authenticate_job(
```
illusional: A small suggestion to make this clearer as we transition to other clouds

```suggestion
def authenticate_cloud_credentials_in_job(
```

illusional (approved):
LGTM! I left one small suggestion for a method name

cpg_utils/hail.py
illusional: I think this is fine for now, I think Hail Batch is assuming the same path for its Azure key, and yeah we'll have some modifications to make when we support calling to GCP from Azure.

cpg_utils/git.py
```diff
@@ -0,0 +1,185 @@
...
+    """
+
+    # Use "set -x" to print the commands for easier debugging.
+    if print_all_statements:
+        job.command('set -x')
+
+    # Note: for private GitHub repos we'd need to use a token to clone.
+    # Any job commands here are evaluated in a bash shell, so user arguments should
+    # be escaped to avoid command injection.
+    job.command(
```
illusional, line 171: I don't think you'd want to quote the organisation and repo_name, just the full URL. Does GH even let you have repo names with characters you'd have to quote?

illusional (approved):

PR #28, opened 2022-04-11
Bump version: 3.3.1 → 3.3.2: Fix use of quote

### LINKED ISSUE #27: Incorrect use of shlex.quote
# Issue

When moving the git functionality from analysis-runner to cpg-utils, I altered how shlex.quote was used. Instead of separately quoting both the organisation and repository, e.g.

```git clone --recurse-submodules https://github.com/'populationgenomics'/'automated-interpretation-pipeline.git'```

The two strings were embedded in a template and the whole entity was `quote`'d. This means that Bash parses the whole command as a single string, i.e. a file or program, and cannot resolve:

```bash
+ 'git clone --recurse-submodules https://github.com/populationgenomics/automated-interpretation-pipeline.git'
/bin/bash: line 16: git clone --recurse-submodules https://github.com/populationgenomics/automated-interpretation-pipeline.git: No such file or directory
```

## Fix

Revert the use of `quote` so the two strings are separately embedded in the template

MattWellie:
Fixes #27 

Reverts use of `quote` to the analysis-runner syntax. See issue for details

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 3.3.1
+current_version = 3.3.2
 commit = True
 tag = False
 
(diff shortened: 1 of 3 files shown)


lgruen (approved):
Thanks! I've also added the "Bump...:" to the PR title, so the squash commit will be named accordingly.

PR #29, opened 2022-04-19
Add get_group_members for supporting GCS access-group cache

illusional:
- Supports https://github.com/populationgenomics/analysis-runner/pull/392
- Intentionally leaving out the `write`
- [ ] Replace get_group_members with access through serverless function


illusional:
Actioned in #127

PR #31, opened 2022-04-25
Add documentation

MattWellie:
Adds documentation relating to the recently transplanted code (git.py from analysis-runner, [link](https://github.com/populationgenomics/analysis-runner/blob/main/analysis_runner/git.py)). Eventually this code could be removed from analysis runner and only hosted here, so user understanding of this will be important.

- includes code examples for new methods in a collection of .MD files
- renames `hail.py` to `hail_batch.py` (this can be reverted, breaking change may not be worth pleasing an overzealous linter)
- new method(s) for identifying the organisation from the repository path
  - available now that we are not hard-restricting to `populationgenomics` 

[x] linter checks pass


.github/workflows/lint.yaml
```diff
@@ -1,5 +1,5 @@
 name: Lint
-on: [push, pull_request]
+on: [push]
```
lgruen, line 2: Will this also still run the linter when a new PR is created?

documentation/hail_batch.md
```diff
@@ -0,0 +1,75 @@
...
+def main():
+    # initiate a hail query back end
+    init_batch()
+
+    # hail object methods
+    ...
+```
+
+When running a Hail Batch workflow using the analysis-runner, the core Driver image will
+have authentication set up, but separate child batches will start with fresh containers.
```
lgruen:
```suggestion
have authentication set up, but separate child batches will start with fresh jobs.
```

lgruen (approved):
Thank you!

lgruen:
Due to the module being renamed, this should be released with a major version bump.

.github/workflows/lint.yaml
MattWellie: yes and no?! There is currently redundancy as the linter runs when you push, then runs again when you make a PR from that push. If you push and it fails, the PR will still show that latest commit as a CI fail without rerunning

MattWellie:
Updated with a major bump, but not releasing yet as unpinned builds will break

documentation/hail_batch.md
MattWellie: I don't think either of these is the exact right message. The message I intended was that although the Driver container has authentication, the same auth also needs to be triggered within containers used for each batch job. E.g. you can run a `gsutil cp` in the driver image, but shouldn't expect that to work in a job container, unless you trigger auth there as well.

.github/workflows/lint.yaml
illusional: Anecdotally I saw no real changed behaviour when I removed it for sample-metadata.

cpg_utils/hail_batch.py
```diff
@@ -30,7 +36,7 @@ def init_batch(**kwargs):
     )
 
 
-def copy_common_env(job: hb.job.Job) -> None:
+def copy_common_env(job: Union[hb.batch.job.BashJob, hb.batch.job.Job]) -> None:
```
illusional: BashJob is a type of Job right?

illusional (approved):

documentation/hail_batch.md
lgruen: Yes, I think that's clear from the current text, thanks!

cpg_utils/hail_batch.py
MattWellie: Yeah, this was a silly change. The type reference for Job was wrong, but doesn't need to be duplicated for subtypes. Altered in latest commit

PR #32, opened 2022-05-09
Port from cpg-pipes: query_command, reference_path

vladsavelyev:
* Add `reference_path` (similar to `image_path`),
* Add `query_command` to submit jobs that run Hail Query code.

### DIFF
--- a/cpg_utils/hail_batch.py
+++ b/cpg_utils/hail_batch.py
@@ -1,12 +1,16 @@
 """Convenience functions related to Hail."""
 
 import asyncio
+import inspect
 import os
-from typing import Optional
+import textwrap
+from pathlib import Path
+from typing import Optional, Union, List
 
 import hail as hl
 import hailtop.batch as hb
-
+from cloudpathlib import CloudPath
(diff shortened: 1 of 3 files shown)


cpg_utils/hail_batch.py
```diff
@@ -226,3 +257,56 @@ def authenticate_cloud_credentials_in_job(
...
+    if hail_billing_project:
+        assert hail_bucket
+        python_cmd += f"""
+import asyncio
+import hail as hl
+asyncio.get_event_loop().run_until_complete(
+    hl.init_batch(
+        default_reference='{default_reference}',
+        billing_project='{hail_billing_project}',
+        remote_tmpdir='{hail_bucket}',
```
MattWellie, line 292: When initiating a hail Batch with a remote tmpdir, does Hail check that the path exists? If so, this CMD would need to come after the GCloud Auth command. I'll have a play locally

MattWellie: the remote_tmpdir is just stored using String manipulation, no problem

MattWellie (approved):

PR #35, opened 2022-05-17
Bump version: 4.1.3 → 4.2.0: Use a config file instead of environment variables

lgruen:
The idea is to use a config file instead of more and more environment variables. This would depend on the analysis-runner uploading a config file as a blob before launching a batch, which is implemented in https://github.com/populationgenomics/analysis-runner/pull/410. We'd only need to set a single environment variable, `CPG_CONFIG_PATH`, to point to the blob.

At the moment, this simply passes through the dictionary returned by the TOML parser. We could instead have a config class with more clearly defined methods that map to the dictionary entries instead of allowing arbitrary strings. But I'm not sure that's adding much value apart from catching potential typos in the variable names when using strings. The current structure has the advantage that it nicely reflects the grouping mechanism in TOML (e.g. `hail` and `workflow` groups).

### DIFF
--- a/cpg_utils/config.py
+++ b/cpg_utils/config.py
@@ -64,7 +64,7 @@ def get_config() -> MutableMapping[str, Any]:
     MutableMapping[str, Any]
     """
 
-    global _config_path, _config
+    global _config
     if _config is None:  # Lazily initialize the config.
         assert (
             _config_path
--- a/cpg_utils/config.py
+++ b/cpg_utils/config.py
@@ -1,0 +1,62 @@
+"""Provides access to config variables."""
+
+import os
+from typing import Any, Dict, Optional, MutableMapping
+from cloudpathlib import AnyPath
+import toml
+
+_config_cache: Dict[str, MutableMapping[str, Any]] = dict()
+
+
+def get_config(config_path: Optional[str] = None) -> MutableMapping[str, Any]:
+    """Returns a configuration dictionary.
+
+    Examples
+    --------
+    Here's a typical configuration file in TOML format:
+
+    [hail]
+    billing_project = "tob-wgs"
+    bucket = "cpg-tob-wgs-hail"
+
+    [workflow]
+    access_level = "test"
+    dataset = "tob-wgs"
+    dataset_gcp_project = "tob-wgs"
+    driver_image = "australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:36c6d4548ef347f14fd34a5b58908057effcde82-hail-ad1fc0e2a30f67855aee84ae9adabc3f3135bd47"
+    image_registry_prefix = "australia-southeast1-docker.pkg.dev/cpg-common/images"
+    reference_prefix = "gs://cpg-reference"
+    output_prefix = "plasma/chr22/v6"
+
+    Notes
+    -----
+    Caches the result based on the config path alone.
+
+    Parameters
+    ----------
+    config_path: str, optional
+        A cloudpathlib-compatible path to a TOML file containing the configuration. If
+        this parameter not provided, the CPG_CONFIG_PATH environment variable is used
+        instead.
+
+    Returns
+    -------
+    dict
+    """
+
+    if config_path is None:
+        config_path = os.getenv('CPG_CONFIG_PATH')
+    assert config_path
+
+    cached = _config_cache.get(config_path)
+    if cached is not None:
+        return cached
+
+    with AnyPath(config_path).open() as f:
+        config_str = f.read()
+
+    # Print the config content, which is helpful for debugging.
+    print(f'Configuration at {config_path}:\n{config_str}')
+    config = _config_cache[config_path] = toml.loads(config_str)
+
+    return config
--- a/cpg_utils/config.py
+++ b/cpg_utils/config.py
@@ -41,7 +41,7 @@ def get_config(config_path: Optional[str] = None) -> MutableMapping[str, Any]:
 
     Returns
     -------
-    dict
+    MutableMapping[str, Any]
     """
 
     if config_path is None:
--- a/cpg_utils/config.py
+++ b/cpg_utils/config.py
@@ -36,8 +36,8 @@ def get_config(config_path: Optional[str] = None) -> MutableMapping[str, Any]:
     ----------
     config_path: str, optional
         A cloudpathlib-compatible path to a TOML file containing the configuration. If
-        this parameter not provided, the CPG_CONFIG_PATH environment variable is used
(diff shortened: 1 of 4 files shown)


cpg_utils/config.py
```diff
@@ -0,0 +1,62 @@
...
+
+import os
+from typing import Any, Dict, Optional, MutableMapping
+from cloudpathlib import AnyPath
+import toml
+
+_config_cache: Dict[str, MutableMapping[str, Any]] = dict()
+
+
+def get_config(config_path: Optional[str] = None) -> MutableMapping[str, Any]:
```
vladsavelyev: [nitpicking] 
Because we know what config structure to expect, we could return a value derived from [TypedDict](https://peps.python.org/pep-0589/) for more strict typing.

cpg_utils/hail_batch.py
```diff
@@ -12,29 +12,30 @@
...
 from cloudpathlib.anypath import to_anypath
 
+from config import get_config
+
 
 GCLOUD_AUTH_COMMAND = (
     'gcloud -q auth activate-service-account --key-file=/gsa-key/key.json'
 )
 
+_config = get_config()
```
vladsavelyev: This means that users should make sure to set `CPG_CONFIG_PATH` before they import cpg_utils. Which makes it 
difficult to allow external users to configure their runs with command line parameters alone, without environment variables. Maybe it's okay, as long as we document this clearly in the readme. They'll anyway have to configure somehow paths to references, images, inputs, Hail, etc. But ideally it should be possible both through configs and command line, e.g. with [click-config](https://github.com/EverythingMe/click-config). I don't have good ideas though.

cpg_utils/config.py
lgruen: `MutableMapping[str, Any]` is what the [`toml` parser](https://github.com/uiri/toml#api-reference) returns. Is there a simple way to convert that to a `TypedDict`?

cpg_utils/hail_batch.py
```diff
@@ -70,20 +62,17 @@ def copy_common_env(job: hb.batch.job.Job) -> None:
...
 
-    If `hail_bucket` is not specified explicitly, requires the HAIL_BUCKET environment variable to be set."""
-
-    if not hail_bucket:
-        hail_bucket = os.getenv('HAIL_BUCKET')
-        assert hail_bucket
-    return f'gs://{hail_bucket}/batch-tmp'
+    If `hail_bucket` is not specified explicitly, requires the `hail/bucket` config variable to be set.
+    """
+    return f'gs://{hail_bucket or _config["hail"]["bucket"]}/batch-tmp'
```
MattWellie: I'm not sure that this is the best place for this, but some checking prior to reaching for the values could be useful?

```suggestion
    bucket = hail_bucket or _config.get('hail', {}).get('bucket')
    assert bucket, f'hail_bucket name was not set by argument or configuration'
    return f'gs://{bucket}/batch-tmp'
```

cpg_utils/config.py
vladsavelyev: I'd guess a simple constructor call like `MyTypedDict(d)`, but will double check!

Even without TypedDict, I wouldn't return `MutableMapping[str, Any]`, but rather convert toml parser result to a `dict` if necessary and change the return type to simple dict: `Dict[str, str]`, to avoid confusion. But again, really nit picking.

lgruen: It would still have to be a `Dict[str, Any]` though, as you can have integer, lists, etc. as values in TOML. If you think `Dict[str, Any]` is better than `MutableMapping[str, Any]`, I'm happy to change this (don't have a strong opinion myself).

cpg_utils/hail_batch.py
lgruen: Good point, thanks! I've added a `set_config_path` method and don't call `get_config()` at the `hail_batch` module scope anymore. Do you think that's better?

Please let me know if there's a way to make pylint happy without disabling the warnings for the globals that are used for lazy initialization.

cpg_utils/hail_batch.py
lgruen: Thanks, that's indeed much better!

MattWellie: maybe an access method somewhere...

```
def _value_from_config(*args):
    assert args, '_value_from_config requires index arguments'
    _lookup = deepcopy(conf)
    for key in args:
        if key not in _lookup:
            raise Exception(f'key {key} was not present in the config dictionary. Requested index: {"::".join(args)}')
        _lookup = _lookup.get(key)

    return _lookup

conf = {'foo': {'foo2':'bar'}}
print(_value_from_config('foo', 'foo2')). # bar
print(_value_from_config('foo', 'bar'))
# Exception: key bar was not present in the config dictionary. Requested index: foo::bar
```

lgruen: While I like that idea, wouldn't a more generic solution be to use something like https://pypi.org/project/traceback-with-variables/ for the whole program?

MattWellie (approved):
Looks good, I can definitely see the benefits. Will the default environment configuration be baked into the Analysis-Runner default image, or in a GCP reference bucket?

cpg_utils/config.py
vladsavelyev: toml actually doesn't work well with TypedDict: https://github.com/uiri/toml/issues/391, so scratch my suggestion :) The idea was to have something like that, to strictly type-check the config:

```py
class ConfigDict(TypedDict, total=False):
    """
    Job attributes specification.
    """

    billing_project: str
    bucket: str
    access_level: str
    dataset: str
    dataset_gcp_project: str
    driver_image: str
    image_registry_prefix: str
    reference_prefix: str
    output_prefix: str


def get_config() -> ConfigDict:
    ...
    _config = ConfigDict(toml.loads(config_str))
    return _config
```

vladsavelyev (approved):

cpg_utils/hail_batch.py
vladsavelyev: I don't think there is a way to make it happy, global variables are just generally discouraged. But an alternative would be an explicit singleton class, which is also too wordy and not very pythonic.

lgruen:
> Looks good, I can definitely see the benefits. Will the default environment configuration be baked into the Analysis-Runner default image, or in a GCP reference bucket?

Thanks! We can't bake it into the image, as some variables (like the dataset or the output location) depend on the specific run. That's why https://github.com/populationgenomics/analysis-runner/pull/410 will copy a unique config blob for each batch to `gs://cpg-config`.

PR #36, opened 2022-05-18
image_path: handle trailing slash in CPG_IMAGE_REGISTRY_PREFIX

vladsavelyev:
To be more robust to e.g.

```
CPG_IMAGE_REGISTRY_PREFIX=australia-southeast1-docker.pkg.dev/cpg-common/images/
```


cpg_utils/hail_batch.py
```diff
@@ -295,7 +295,7 @@ def query_command(
     )
 )
 """
-        python_cmd += f"""
+    python_cmd += f"""
```
vladsavelyev, line 298: That's intentional :) Just another minor fix

vivbak (approved):
🙏 Thanks!

PR #37, opened 2022-05-18
Add web_url

vladsavelyev:
Add function to get CPG web URL. URL template is parametrised by an environment variable, similar to other CPG paramers. The variable is added into the server AR code: https://github.com/populationgenomics/analysis-runner/pull/411

~Make sure to merge https://github.com/populationgenomics/cpg-utils/pull/36 first~

### DIFF
--- a/cpg_utils/hail_batch.py
+++ b/cpg_utils/hail_batch.py
@@ -17,6 +17,8 @@ GCLOUD_AUTH_COMMAND = (
     'gcloud -q auth activate-service-account --key-file=/gsa-key/key.json'
 )
 
+WEB_URL_FORMAT = 'https://{namespace}-web.populationgenomics.org.au/{dataset}'
+
 
 def init_batch(**kwargs):
     """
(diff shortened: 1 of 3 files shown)


vivbak (approved):
🙏  Thanks!!

cpg_utils/hail_batch.py
```diff
@@ -17,6 +17,8 @@
     'gcloud -q auth activate-service-account --key-file=/gsa-key/key.json'
 )
 
+WEB_URL_FORMAT = 'https://{namespace}-web.populationgenomics.org.au/{dataset}'
```
lgruen: Should we make this configurable or will `web_url` only be called for CPG-internal implementations?

lgruen (changes requested):

cpg_utils/hail_batch.py
vladsavelyev: Currently in the cpg-pipes, `CpgStorageProvider` uses the `web_url` function from cpg_utils, and the abstract one doesn't have a default implementation.

But if we are getting rid of those abstractions and relying entirely on the toml config to configure the environment, then we should somehow explain that the WEB URL value should be parametrised by namespace and dataset. It's straightforward to put such requirements with abstract classes, but a bit more error-prone with a string template like that. But could add `web_url_template` into the toml config with a comment.

vladsavelyev: Added a commit to parametrise it with an environment variable

lgruen (approved):

PR #40, opened 2022-05-20
Parametrise path scheme in `dataset_path`

vladsavelyev:
Parametrise the cloud provider path scheme for `dataset_path`. For Azure, assuming the Hail's `hail-az` scheme.

Added: 
* `cloudpath_hail_az.py` module is an extension for `cloudpathlib` to support the `hail-az` scheme with `CloudPath`.
* Into `hail_bactch.py`: 
   * An abstract class for path protocol along with 2 implementations for `gs` and `hail-az` correspondingly.
   * Default path scheme is `gs`, and it can be parametrise in the config through `workflow/path_scheme`.
   * Explicitly parametrise `dataset_path()` with `dataset`, `access_level`, and `path_scheme`.
   * Add `Path` and `to_path` shortcuts are used everywhere in `cpg-pipes`, would be nice to have them in `cpg-utils` to aid moving code from one repository into another.

### DIFF
--- a/cpg_utils/cloudpath_hail_az.py
+++ b/cpg_utils/cloudpath_hail_az.py
@@ -1,0 +1,107 @@
+"""
+Extend cloudpathlib Azure implementation to support hail-az:// scheme.
+Inspired by https://github.com/drivendataorg/cloudpathlib/issues/157
+"""
+
+import re
+from typing import Union, Optional
+from urllib.parse import urlparse
+
+from cloudpathlib import AzureBlobClient, AzureBlobPath
+from cloudpathlib.client import register_client_class
+from cloudpathlib.cloudpath import register_path_class, CloudPath
+from cloudpathlib.exceptions import InvalidPrefixError
+
+
+@register_client_class('hail-az')
+class HailAzureBlobClient(AzureBlobClient):
+    """
+    Client is identical to the base Azure implementation.
+    """
+
+
+@register_path_class('hail-az')
+class HailAzureBlobPath(AzureBlobPath):
+    """
+    Extending Path implementation to support hail-az:// and https:// schemes.
+    >>> CloudPath('hail-az://myaccount/mycontainer/tmp')
+    HailAzureBlobPath('hail-az://myaccount/mycontainer/tmp')
+    >>> CloudPath('https://myaccount.blob.core.windows.net/mycontainer/tmp')
+    HailAzureBlobPath('hail-az://myaccount/mycontainer/tmp')
+    """
+
+    cloud_prefix: str = 'hail-az://'
+    client: 'HailAzureBlobClient'
+
+    def __init__(
+        self,
+        cloud_path: Union[str, CloudPath],
+        client: Optional[AzureBlobClient] = None,
+        token: Optional[str] = None,
+    ):
+        if isinstance(cloud_path, str):
+            parsed = urlparse(cloud_path)
+            m = re.match(
+                r'(?P<account>[a-z0-9]+)(\.(?P<type>blob|dfs)(\.core\.windows\.net)?)?',
+                parsed.netloc,
+                flags=re.IGNORECASE,
+            )
+            if m is None:
+                raise ValueError(f"Bad Azure path '{cloud_path}'")
+            account = m.group('account')
+            fstype = m.group('type') or 'blob'
+            account_url = f'https://{account}.{fstype}.core.windows.net/'
+            optional_type = '' if fstype == 'blob' else '.' + fstype
+            cloud_path = (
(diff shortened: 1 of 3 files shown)


cpg_utils/hail_batch.py
```diff
@@ -64,6 +72,9 @@ def init_batch(**kwargs):
     kwargs : keyword arguments
         Forwarded directly to `hl.init_batch`.
     """
+
+    billing_project = os.getenv('HAIL_BILLING_PROJECT')
```
lgruen: Looks like this didn't get resolved properly when merging with HEAD?

lgruen (changes requested):

cpg_utils/hail_batch.py
vladsavelyev: Oh. Yeah, merge artefact. Fixed.

vladsavelyev: There are few more merge artefacts, will address them now

cpg_utils/cloudpath_hail_az.py
```diff
@@ -0,0 +1,107 @@
...
+    ):
+        if isinstance(cloud_path, str):
+            parsed = urlparse(cloud_path)
+            m = re.match(
+                r'(?P<account>[a-z0-9]+)(\.(?P<type>blob|dfs)(\.core\.windows\.net)?)?',
+                parsed.netloc,
+                flags=re.IGNORECASE,
+            )
+            if m is None:
+                raise ValueError(f"Bad Azure path '{cloud_path}'")
```
lgruen:
```suggestion
                raise ValueError(f'Bad Azure path "{cloud_path}"')
```

cpg_utils/hail_batch.py
```diff
@@ -81,11 +54,21 @@ def copy_common_env(job: hb.batch.job.Job) -> None:
...
     The environment variable values are extracted from the current process and
-    copied to the environment dictionary of the given Hail Batch job.
-    """
-    # If possible, please don't add new environment variables here, but instead add
-    # config variables.
-    for key in ('CPG_CONFIG_PATH',):
+    copied to the environment dictionary of the given Hail Batch job."""
+
+    for key in (
+        'CPG_ACCESS_LEVEL',
```
lgruen: Looks like there's still some merge remnants.

lgruen (changes requested):

cpg_utils/hail_batch.py
vladsavelyev: Somehow screwed up merge with main. Should be good now!

cpg_utils/hail_batch.py
```diff
@@ -101,7 +117,73 @@ def remote_tmpdir(hail_bucket: Optional[str] = None) -> str:
...
+        """Build full path from prefix and suffix"""
+
+    @staticmethod
+    def parse(val: str) -> 'PathScheme':
+        """Parse subclass name from string"""
+        if val == 'gs':
+            return GSPathScheme()
+        if val in ['az', 'hail-az']:
+            return AzurePathScheme()
+        raise ValueError(f'Unsupported path format: {val}. Available: gs, hail-az')
```
lgruen:
```suggestion
        match val:
            case 'gs':
                return GSPathScheme()
            case 'hail-az':
                return AzurePathScheme()
            case _:
                raise ValueError(f'Unsupported path format: {val}. Available: gs, hail-az')
```

cpg_utils/hail_batch.py
```diff
@@ -101,7 +117,73 @@ def remote_tmpdir(hail_bucket: Optional[str] = None) -> str:
...
+
+class AzurePathScheme(PathScheme):
+    """
+    Azure Blob Storage path scheme, following the Hail Batch hail-az format.
+    """
+
+    def __init__(self, account: Optional[str] = 'cpg'):
+        config = get_config()
+        self.scheme = 'hail-az'
+        self.account = config['workflow'].get('azure_account', account)
```
lgruen, line 171:
```suggestion
        self.scheme = 'hail-az'
        self.account = get_config()['workflow'].get('azure_account', account)
```

lgruen (approved):

cpg_utils/hail_batch.py
vladsavelyev: Any reason why it's better?

cpg_utils/cloudpath_hail_az.py
vladsavelyev: Thanks!

cpg_utils/hail_batch.py
lgruen: Just that `config` isn't reused anywhere, so it's just one less variable to keep track of :)

PR #42, opened 2022-05-27
Update pre-commit hooks

vladsavelyev:


lgruen (approved):

lgruen:
Not sure why the linter is complaining about `toml`, as it's mentioned here?
https://github.com/populationgenomics/cpg-utils/blob/a33065555154fabe3558f49e8d24e774c90aa475/setup.py#L19

PR #45, opened 2022-05-30
Bump version: 4.2.2 → 4.2.3: Use dict instead of MutableMapping

lgruen:
Should not result in any functional change.

### DIFF
--- a/cpg_utils/config.py
+++ b/cpg_utils/config.py
@@ -1,14 +1,14 @@
 """Provides access to config variables."""
 
 import os
-from typing import Any, MutableMapping, Optional
+from typing import Optional
 from cloudpathlib import AnyPath
 import toml
 
 # We use these globals for lazy initialization, but pylint doesn't like that.
 # pylint: disable=global-statement, invalid-name
 _config_path = os.getenv('CPG_CONFIG_PATH')  # See set_config_path.
-_config: Optional[MutableMapping[str, Any]] = None  # Cached config, initialized lazily.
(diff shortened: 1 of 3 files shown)


MattWellie (approved):

cpg_utils/config.py
```diff
@@ -1,14 +1,14 @@
...
-from typing import Any, MutableMapping, Optional
+from typing import Optional
 from cloudpathlib import AnyPath
 import toml
 
 # We use these globals for lazy initialization, but pylint doesn't like that.
 # pylint: disable=global-statement, invalid-name
 _config_path = os.getenv('CPG_CONFIG_PATH')  # See set_config_path.
-_config: Optional[MutableMapping[str, Any]] = None  # Cached config, initialized lazily.
+_config: Optional[dict] = None  # Cached config, initialized lazily.
```
illusional, line 11: I prefer `Dict` from typing, then you can annotate it like `Dict[str, Any]`, but this is fine too

illusional (approved):
Only nitpick about which `dict / Dict` to type annotate with

cpg_utils/config.py
vladsavelyev: @illusional you can use `dict[str, Any]`

vladsavelyev (approved):

cpg_utils/config.py
illusional: Ah that's a Python 3.10 thing I see

PR #46, opened 2022-06-02
Bump version: 4.3.0 → 4.3.1: Add config default `config-template.toml`, get references and images by key

vladsavelyev:
For this PR to work: https://github.com/populationgenomics/production-pipelines/pull/60

* load config baseline from `config-template.toml`
* use `key` in `reference_path()` and `image_path()` instead of suffix to load from config, support sections
* add `genome_build()`,
* add `fasta_res_group()`.
* add `write_config()` function to write config to a new location and set it as active


vladsavelyev:
Ah, probably adding the "Namespace" class should go as another PR. 
UPD: done https://github.com/populationgenomics/cpg-utils/pull/47

cpg_utils/hail_batch.py
```diff
@@ -101,7 +85,7 @@ def copy_common_env(job: hb.batch.job.Job) -> None:
     """
     # If possible, please don't add new environment variables here, but instead add
     # config variables.
-    for key in ('CPG_CONFIG_PATH',):
+    for key in ('CPG_CONFIG_PATH', 'HAIL_TOKEN'):
```
lgruen: Was the addition of `HAIL_TOKEN` here deliberate?

cpg_utils/hail_batch.py
```diff
@@ -181,8 +165,37 @@ def full_path(self, prefix: str, suffix: str) -> str:
...
+            Namespace.TEST: ['test'],
+        }.items():
+            if str_val in str_vals:
+                return val
+        raise ValueError(f'Cannot parse namespace or access level {str_val}')
+
+
 def dataset_path(
-    suffix: str,
+    suffix: str = '',
```
lgruen: Is it useful to have an empty default value here? (Also below.)

cpg_utils/hail_batch.py
```diff
@@ -337,16 +350,19 @@ def image_path(suffix: str) -> str:
     Parameters
     ----------
     suffix : str
-        Describes the location within the registry.
+        Describes the location within the registry, or key in the `images` config
```
lgruen: Should we always require that images are specified in the config, i.e. only support keys here?

cpg_utils/hail_batch.py
```diff
@@ -361,14 +377,54 @@ def reference_path(suffix: str) -> Union[CloudPath, Path]:
...
+    suffix = suffix.strip('/')
+    return prefix / suffix
+
+
+def genome_build() -> str:
+    """Return the genome build name"""
+    return get_config()['references'].get('genome_build', 'GRCh38')
+
+
+def fasta_res_group(b, indices: list | None = None):
```
lgruen, line 407: Should this have "reference" in the method name to make it more clear what this is used for?

lgruen (changes requested):

cpg_utils/hail_batch.py
vladsavelyev: Removed

cpg_utils/hail_batch.py
vladsavelyev: Sometimes it's useful to get the prefix, cast it to Path, and pass it around as a typed Path object, e.g. https://github.com/populationgenomics/production-pipelines/blob/initial/scripts/make_seqr_metadata.py#L62

vladsavelyev: Not necessary for image_path() though, I'll remove

cpg_utils/hail_batch.py
vladsavelyev: Sure. Will have to send another PR in `automated-interpretation-pipeline` that calls image_path and reference_path

cpg_utils/hail_batch.py
vladsavelyev: I don't know, I think I'd leave it.

Sorry for pushing back, but that's a good example of a trivial comment/suggestion that I'd love if we try to avoid in PRs!

cpg_utils/hail_batch.py
vladsavelyev: Hmm, in this case `automated-interpretation-pipeline` should somehow know `config-template.toml` from cpg-pipes

vladsavelyev: Should we move `config-template.toml` here?

vladsavelyev: https://github.com/populationgenomics/automated-interpretation-pipeline/pull/43

lgruen: Yes, I think moving it here is fine -- we wanted to merge `cpg-pipes` and `cpg-utils` anyway, right?

cpg_utils/hail_batch.py
lgruen: Of course you're free to disregard the suggestion, but I think commenting on unclear naming isn't something that we should refrain from in PRs.

Just reading this function name (imagine you're new to this module!), it's pretty unclear what this is used for.

cpg_utils/hail_batch.py
```diff
@@ -101,7 +85,7 @@ def copy_common_env(job: hb.batch.job.Job) -> None:
     """
     # If possible, please don't add new environment variables here, but instead add
     # config variables.
-    for key in ('CPG_CONFIG_PATH',):
+    for key in 'CPG_CONFIG_PATH':
```
lgruen: This will iterate over the string, as it's not a tuple anymore.

cpg_utils/hail_batch.py
lgruen: What I tried to imply is that this is probably not the _default_ use case, hence it shouldn't be the default value. You can always get the same behavior by explicitly passing the empty string.

lgruen (approved):
Added a few more comments, but only the one regarding `for key in 'CPG_CONFIG_PATH':` is critical to address.

cpg_utils/hail_batch.py
vladsavelyev: Good catch, thanks!

lgruen (approved):

PR #47, opened 2022-06-02
Add Namespace enum

vladsavelyev:
Moving the Namespace enum used in cpg-pipes into cpg-utils.

Helps to make namespace labels more statically typed, and also to map access levels to namespaces.

### DIFF
--- a/cpg_utils/hail_batch.py
+++ b/cpg_utils/hail_batch.py
@@ -4,6 +4,7 @@ import asyncio
 import inspect
 import os
 import textwrap
+from enum import Enum
 from typing import Optional, List
 from abc import ABC, abstractmethod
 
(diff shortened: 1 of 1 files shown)


cpg_utils/hail_batch.py
```diff
@@ -164,6 +165,35 @@ def full_path(self, prefix: str, suffix: str) -> str:
...
+    https://github.com/populationgenomics/team-docs/tree/main/storage_policies#main-vs-test
+    """
+
+    MAIN = 'main'
+    TEST = 'test'
+
+    @staticmethod
+    def parse(str_val: str) -> 'Namespace':
+        """
+        Parse value from a string.
```
lgruen: This isn't so much parsing an enum string representation, but mapping an access level to a namespace, right? Might be worth clarifying that in the method name / documentation.

lgruen (approved):

cpg_utils/hail_batch.py
vladsavelyev: Renamed to from_access_level

PR #48, opened 2022-06-06
Bump version: 4.2.3 → 4.3.0: Support comma-separated list of configs

lgruen:


cpg_utils/config.py
```diff
@@ -32,7 +32,38 @@ def set_config_path(config_path: str) -> None:
...
+
+        _config = read_config(_config_path)
+
+        # Print the config content, which is helpful for debugging.
+        print(f'Configuration at {_config_path}:\n{toml.dumps(_config)}')
+
+    return _config
+
+
+def read_config(config_path: str) -> dict:
```
vladsavelyev: Should it take a list, and the usage would be `read_config(_config_path.split(','))`?

vladsavelyev (approved):

cpg_utils/config.py
lgruen: Yes, thanks, that's much better!

PR #51, opened 2022-06-08
Bump version: 4.3.3 → 4.3.4: Use driver image instead of deprecated sm-api image

vladsavelyev:


cpg_utils/config-template.toml
```diff
@@ -150,7 +150,7 @@ dry_run = false
 # Docker image URLs. Can be absolute or relative to `workflow.image_registry_prefix`.
 gatk = 'gatk:4.2.6.1'
 bcftools = 'bcftools:1.10.2--h4f4756c_2'
-sm-api = 'sm-api:4.0.0'
+sm-api = 'australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:7d00c4871b2e96f50bae208e4184c3c4789a2fa4-hail-83056327f288917537531475ba475287b413db1c'
```
illusional, line 153: Do you have to update did everytime the driver image updates? Feels very manual

illusional (approved):

cpg_utils/config-template.toml
vladsavelyev: Yep, like other image versions. Not sure what's the best approach here

PR #52, opened 2022-06-09
 Bump version: 4.3.4 → 4.3.5: Add `fail_gracefully` flag to `read_secret`

vladsavelyev:
* With the default `fail_gracefully=True`, current behaviour is kept, will only additionally print the exception.
* With `fail_gracefully=False`, will raise the exception.


lgruen (approved):
🙇

PR #53, opened 2022-06-10
Allow absolute image paths in config

vladsavelyev:
* The driver image is not in `cpg-common`, so modifying `image_path()` to allow absolute image paths in the config (which might also be useful in general?). Alternatively we could push the driver image to `cpg-common` as well. Also not sure about azure prefix.

* Slight fixes of the `config-template`.

### DIFF
--- a/cpg_utils/config-template.toml
+++ b/cpg_utils/config-template.toml
@@ -30,9 +30,8 @@ access_level = 'standard'
 #last_stage =
 
 # Map of stages to lists of samples, to skip for specific stages
-# skip_samples_stages = {
-#   VerifyBamId = ['CPG13409']  # https://batch.hail.populationgenomics.org.au/batches/56236/jobs/372
-# }
+#[workflow.skip_samples_stages]
+#VerifyBamId = ['CPG13409']  # https://batch.hail.populationgenomics.org.au/batches/56236/jobs/372
 
 # Name of the pipeline (to prefix output paths)
 #name =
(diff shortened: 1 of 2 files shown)


cpg_utils/config-template.toml
```diff
@@ -150,20 +146,20 @@ dry_run = false
...
 picard_samtools = 'picard_samtools:v0'
 somalier = 'somalier:v0.2.15'
 peddy = 'peddy:v0'
-vep ='vep:105'
+vep = 'vep:105'
 verify-bam-id = 'verify-bam-id:1.0.1'
 multiqc = 'multiqc:v1.12'
 fastqc = 'fastqc:v0.11.9_cv8'
 hail = 'australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:7d00c4871b2e96f50bae208e4184c3c4789a2fa4-hail-83056327f288917537531475ba475287b413db1c'
+sm-api = 'australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:7d00c4871b2e96f50bae208e4184c3c4789a2fa4-hail-83056327f288917537531475ba475287b413db1c'
```
lgruen, line 162: This kind of defeats the purpose of the image prefix, for which the main motivation was that people could copy / rebuild the images themselves in a local repository and switch everything over by only setting a new prefix.

It might make sense to think about how we want to make this set of images available to others that want to run the same pipelines, e.g. the AIP on the Broad or Microsoft side. Would we provide another script that populates a Docker image registry?

Should images that fulfill the `hail` and `sm-api` functionality be part of our [images repository](https://github.com/populationgenomics/images), for example? That way, others might have an easier time to rebuild the images.

@illusional What are your thoughts on this?

lgruen (changes requested):

cpg_utils/config-template.toml
illusional: Seems reasonable to me! I turned off the sm-api because it was causing build issues, the functionality is covered by the driver_image and I also incorrectly assumed it wasn't being used.

I'm happy to contribute a command to build the SM-API docker image (it's just a plain old python image with a pip install now).

PR #54, opened 2022-06-14
upload pathing change

MattWellie:
I... think this is a correction?

When using dataset_path to create an upload path (i.e. cpg-validation-main-upload), this method currently creates `cpg-validation-upload` instead. 

Based on the [storage policy document](https://github.com/populationgenomics/team-docs/tree/main/storage_policies#upload-gscpg-dataset-maintest-upload), the `-upload` suffix should follow `test/main`, not replace it

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 4.3.5
+current_version = 4.3.6
 commit = True
 tag = False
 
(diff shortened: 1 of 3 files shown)


lgruen (approved):
Thank you! There used to be only one `upload` bucket, but we added "namespaced" ones later.

lgruen:
Do you also want to add a version bump commit?

PR #56, opened 2022-06-24
Add happy vcfeval to images

MattWellie:
Adds image reference to the Hap.py & vcfeval docker image

refs https://console.cloud.google.com/artifacts/docker/cpg-common/australia-southeast1/images/happy-vcfeval?project=cpg-common

### DIFF
--- a/cpg_utils/config-template.toml
+++ b/cpg_utils/config-template.toml
@@ -163,6 +163,7 @@ vep ='vep:105'
 verify-bam-id = 'verify-bam-id:1.0.1'
 multiqc = 'multiqc:v1.12'
 fastqc = 'fastqc:v0.11.9_cv8'
+happy-vcfeval = 'happy-vcfeval:1.0'
 hail = 'australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:7d00c4871b2e96f50bae208e4184c3c4789a2fa4-hail-83056327f288917537531475ba475287b413db1c'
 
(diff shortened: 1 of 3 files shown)


cpg_utils/config-template.toml
```diff
@@ -163,6 +163,7 @@ vep ='vep:105'
 verify-bam-id = 'verify-bam-id:1.0.1'
 multiqc = 'multiqc:v1.12'
 fastqc = 'fastqc:v0.11.9_cv8'
+happy-vcfeval = 'happy-vcfeval:1.0'
```
lgruen, line 166: Is this something we want in the generic template or an addition in the AIP configuration?

Note that you can specify multiple config files and they get merged in order.

lgruen (changes requested):

cpg_utils/config-template.toml
MattWellie: My main thought was that the image will initially be used in working out how to run the validation, but as the validation stack can be dropped into any cohort, the QC stage may become a standard post-joint-call main pipeline job in future.

If there's substantial benefit in normalising variants, vcfeval could also become a non-validation part of the pipeline.

Happy to use a minimal config to update

lgruen: Sounds good to me if you think this is more generally useful in the future!

lgruen (approved):

PR #62, opened 2022-08-11
Drop python 3.10 specific syntax for dataproc

vladsavelyev:


lgruen (approved):
I wish we could build a Dataproc image with Python 3.10, but it's probably not easy.

PR #68, opened 2022-08-29
VEP 104

vladsavelyev:
Adjusting pointers in `images` and `references` to use VEP 104 by default.

FYI @MattWellie and @KatalinaBobowik it would break IAP and the VEP dataproc script in `ancestry`, but I'll follow up with corresponding PRs right after testing this change.

UPD: [context](https://centrepopgen.slack.com/archives/C018KFBCR1C/p1661488085779609?thread_ts=1658208231.938539&cid=C018KFBCR1C). Happy to discuss if we want to support different versions for different domains. Or we could treat TOB as one-off, manually pin for one specific run, but use the latest and greatest in the default configs.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 4.5.2
+current_version = 4.5.3
 commit = True
 tag = False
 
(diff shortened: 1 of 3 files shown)


MattWellie:
What is the reason to use 104.3 over 105?

vladsavelyev:
@MattWellie this is the context:
https://centrepopgen.slack.com/archives/C018KFBCR1C/p1661488085779609?thread_ts=1658208231.938539&cid=C018KFBCR1C
But I guess as Leo pointed, versions don't need to match between RD and popgen domains, so we could think of the best ways to support multiple versions, I guess.

MattWellie:
@vladsavelyev  I've got 3 more AIP runs to set off today, but none are being re-annotated so happy to approve this now

MattWellie (approved):

PR #69, opened 2022-08-29
Use VEP 105 for RD

vladsavelyev:
As a one-off for TOB in the populationgenomics domain, we use the `hl.vep()` driven by the dataproc backend to annotate, and there we can specifically pin to an earlier version. But for the batch-backend vep workfllow that is used for the RD domain, can keep the default to 105.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 4.5.3
+current_version = 4.5.4
 commit = True
 tag = False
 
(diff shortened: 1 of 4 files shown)


MattWellie:
Is there any situation where we would use 104 again, or is the annotation data going to be deleted? I was thinking that `image('vep:105')` and `image('vep:104')` can both co-exist

vladsavelyev:
The dataproc backend doesn't call `image_path()` explicitly, so no need to support it here. We can think of a more generic ways to support multiple versions with `image_path` and `reference_path`, but that's down the road! Don't want to hardcode it now :)

MattWellie (approved):
LGTM

PR #73, opened 2022-09-07
Support filtering samples by batch number

vladsavelyev:
Useful for large-cohort datasets.

Batch IDs are set in the config, similar to `only_samples`:

```toml
[workflow]
only_batches = [1, 2]
```

### DIFF
--- a/cpg_utils/workflows/inputs.py
+++ b/cpg_utils/workflows/inputs.py
@@ -62,7 +62,9 @@ def create_cohort() -> Cohort:
         return cohort
 
     if sequencing_type := get_config()['workflow'].get('sequencing_type'):
-        _populate_alignment_inputs(cohort, sequencing_type)
+        only_batches = get_config()['workflow'].get('only_batches', [])
+        only_batches = [str(b) for b in only_batches]
+        _populate_alignment_inputs(cohort, sequencing_type, only_batches)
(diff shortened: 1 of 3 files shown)


cpg_utils/workflows/inputs.py
```diff
@@ -161,6 +157,24 @@ def _populate_alignment_inputs(
     for sample in cohort.get_samples():
         for d in found_seqs_by_sid.get(sample.id, []):
             seq = Sequence.parse(d, check_existence=check_existence)
+            if only_batches:
+                if 'batch' not in seq.meta:
```
illusional, line 161: Should probably come up with a list of these "patterns" around the place. Or even maybe define them in MM itself to make it clearer how these are used and provided.

vladsavelyev: Agree, makes sense to do this in metamist!

lgruen: It also took me a few seconds to realize that this is about _sample_ batches and completely unrelated to Hail batches. The term "batch number" is unfortunately somewhat overloaded. This becomes even more tricky as we also want to store Hail Batch related information in Metamist, e.g. resource usage.

Can we make this a little clearer by using `sample_batch` or similar instead?

lgruen (changes requested):
(See comment about "batch" naming.)

PR #77, opened 2022-09-16
Bump version: Add ability to override default reference

illusional:


MattWellie (approved):

vladsavelyev:
Just a note that you can set reference in the toml config (that's where `genome_build()` reads it from): 
```sh
[references]
genome_build = 'GRCh38'
```
https://github.com/populationgenomics/cpg-utils/blob/main/cpg_utils/config-template.toml#L150

vladsavelyev:
@MattWellie and @illusional, you can also override reference using the toml config (that's where `genome_build()` reads it from): 
```sh
[references]
genome_build = 'GRCh38'
```
https://github.com/populationgenomics/cpg-utils/blob/main/cpg_utils/config-template.toml#L150

illusional:
Maybe it needs to be configurable on analysis runner CLI? Because currently it's a bit of pain to override then? You'd need to write an additional file, then reexport the environment variable right?

vladsavelyev:
You can pass a config via `--config` of analysis runner:
https://github.com/populationgenomics/analysis-runner/blob/main/analysis_runner/cli_analysisrunner.py#L76-L86

But I'm curious if we ever need to actually override the default reference, i.e. if there is any reason to use GRCh37?

MattWellie:
The real question is will we be supporting `GRCh38` & `39` at some point 🌚

vladsavelyev:
@MattWellie 😆 
GRCh39 will never happen :)

PR #79, opened 2022-09-22
Revert reference override

illusional:
Reverting in favour of adding this line to the analysis-runner invocation. Basically it just asks analysis-runner to override the reference.genome_build parameter. You can pass a file, eg:

`grch37.yoml`
```toml
[reference]
genome_build = "GRCh37"
```

`analysis-runner --config grch37.toml [...]`

You can do this inline with process-substitution too, eg:

```shell
analysis-runner \
    --config <(echo '[references]\ngenome_build = "GRCh37"') \
    [...]
```

### DIFF
--- a/cpg_utils/hail_batch.py
+++ b/cpg_utils/hail_batch.py
@@ -40,9 +40,9 @@ def init_batch(**kwargs):
         return  # already initialised
     dataset = get_config()['workflow']['dataset']
     kwargs.setdefault('token', os.environ.get('HAIL_TOKEN'))
-    kwargs.setdefault('default_reference', genome_build())
     asyncio.get_event_loop().run_until_complete(
         hl.init_batch(
+            default_reference=genome_build(),
             billing_project=get_config()['hail']['billing_project'],
             remote_tmpdir=remote_tmpdir(f'cpg-{dataset}-hail'),
             **kwargs,
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 4.7.1
+current_version = 4.8.0
 commit = True
 tag = False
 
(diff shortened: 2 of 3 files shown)


tiboloic (commented):
I am not sure if there is already a convention at CPG about this, but I would remove any reference to `hg19` and replace by `GRCh37`

illusional:
Ah yeah sorry, this was just me spitballing on how it might look - I'll change the example.

tiboloic (approved):

cpg_utils/hail_batch.py
```diff
@@ -40,9 +40,9 @@ def init_batch(**kwargs):
         return  # already initialised
     dataset = get_config()['workflow']['dataset']
     kwargs.setdefault('token', os.environ.get('HAIL_TOKEN'))
-    kwargs.setdefault('default_reference', genome_build())
     asyncio.get_event_loop().run_until_complete(
         hl.init_batch(
+            default_reference=genome_build(),
             billing_project=get_config()['hail']['billing_project'],
             remote_tmpdir=remote_tmpdir(f'cpg-{dataset}-hail'),
             **kwargs,
```
lgruen, line 48: What happens if `default_reference` is also specified in the `kwargs`? Does the latter take precedence? Or are you trying to forbid that? (This seems to be the only functional change, right?)

lgruen (changes requested):

cpg_utils/hail_batch.py
illusional: I'm reverting to preserve the original behaviour, which in this case which cause a TypeError (`init_batch() got multiple values for keyword argument 'default_reference'`. To be clear, I want to forbid this introduced behaviour of allowing the override of the default reference.

lgruen (approved):

PR #80, opened 2022-09-26
Filter samples according to access level 

vivbak:
Currently a config file that looks like this 

> [workflow]
> dataset = 'dataset'
> datasets = ['dataset']
> access_level = 'test'

Will read/write analysis objects in the `dataset-test` metamist project, but read samples from the `dataset` metamist project. 

This PR suggests behaving consistently in terms of what metamist project to read from when specifying test access-level. 

There may be a use-case for reading samples from the main project in metamist and writing analyses to the corresponding -test project. This PR does not consider that use case lol.  [Perhaps we should instead split access_level into two variables to capture the difference?]

Oh also, get_pedigree needed an updated param name.


lgruen:
Sorry for the fly-by comment, but any chance we could rename the two `dataset` and `datasets` parameters within the `workflow` section, to make the semantics more obvious? Maybe something about output vs inputs?

cpg_utils/workflows/inputs.py
```diff
@@ -207,9 +210,11 @@ def _populate_participants(cohort: Cohort) -> None:
     """
     for dataset in cohort.get_datasets():
         logging.info(f'Reading participants IDs for dataset {dataset}')
+        if get_config()['workflow']['access_level'] == 'test':
+            dataset_name = dataset.name + '-test'
         pid_sid_multi = (
             get_metamist().papi.get_external_participant_id_to_internal_sample_id(
-                dataset.name
+                dataset_name
```
vladsavelyev: It's not a dataset name though after adding the `-test` prefix

cpg_utils/workflows/metamist.py
```diff
@@ -384,11 +384,14 @@ def get_ped_entries(self, dataset: str | None = None) -> list[dict[str, str]]:
...
 
+        if get_config()['workflow']['access_level'] == 'test':
+            dataset += '-test'
+
         families = self.fapi.get_families(dataset)
         family_ids = [family['id'] for family in families]
         ped_entries = self.fapi.get_pedigree(
             internal_family_ids=family_ids,
-            response_type='json',
+            export_type='json',
```
vladsavelyev: What's the difference between `response_type` and `export_type `?

vladsavelyev:
Just a thought, maybe we could be explicit about reading from `-test`?

```
[workflow]
dataset = 'dataset-test'
datasets = ['dataset-test']
access_level = 'test'
```

cpg_utils/workflows/metamist.py
vivbak: There might have been an update or something, but `response_type` isn't the valid param name https://sample-metadata.populationgenomics.org.au/documentation/FamilyApi.md#get_pedigree

cpg_utils/workflows/inputs.py
vivbak: good point, i'll switch to `metamist_proj` I guess?

cpg_utils/workflows/metamist.py
vladsavelyev: Thanks!

cpg_utils/workflows/inputs.py
vladsavelyev: Yep, that's perfect!

vivbak:
I think the only reason I didn't suggest this is because it messed with a bunch of dataset_path functions.

vladsavelyev:
Yes, it would need additional work to implement. But we can leave it as you suggested originally, since you can't read from main metamist projects with a test access level anyway.

vladsavelyev:
Another question though - did you check all places in `metamist.py` and `inputs.py` where we want to add the -test prefix? E.g. _populate_analysis, etc. Maybe we should add a test case for that as well.

vivbak:
Ah, I think I missed a few places because there's a little inconsistency in terms of where the access_level gets checked i.e. sometimes in `inputs.py` and sometimes in `metamist.py`. Because `inputs.py` sometimes calls the endpoints directly e.g. `sample_entries = get_metamist().sapi.get_samples` I think i'll move all of the access level logic out of `metamist.py` and into `inputs.py` if thats ok. 

_Edit_: _Actually leaning toward moving it all to metamist.py_

vladsavelyev:
You can wrap `sample_entries = get_metamist().sapi.get_samples` in `Metamist` and avoid calling endpoints in `inputs.py`, and handle `access_level` in `Metamist`.

vladsavelyev:
And two more things:
1. Please make sure the `-test` projects are actually populated (at least `validation-test` and `acute-care-test` if the latter is used for gatk-sv testing),
2. CI checks are currently failing. Maybe install the hooks locally as well

vladsavelyev (changes requested):
(See two comments above)

vladsavelyev:
> You can wrap sample_entries = get_metamist().sapi.get_samples in Metamist and avoid calling endpoints in inputs.py, and handle access_level in Metamist.

I've done that on my side! I'll merge the PR, but please still populate the `-test` projects

vladsavelyev (approved):

PR #81, opened 2022-09-27
Bump version: 4.8.0 → 4.8.1: Attempt to fetch default credentials first

lgruen:
Context: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1664169251815529?thread_ts=1663970565.119819&cid=C030X7WGFCL

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 4.8.0
+current_version = 4.8.1
 commit = True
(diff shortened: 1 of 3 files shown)


illusional (approved):
Don't need the logger anymore?

lgruen:
> Don't need the logger anymore?

That was mainly to suppress the spam that's logged when the GCE path gets executed for human accounts. Since we swapped the order now, that shouldn't be a problem anymore.

PR #89, opened 2022-10-04
datasets -> only_datasets

vivbak:
Following comment here https://github.com/populationgenomics/cpg-utils/pull/80 ~~and in line with only_samples convention~~

### DIFF
--- a/cpg_utils/config-template.toml
+++ b/cpg_utils/config-template.toml
@@ -9,7 +9,7 @@ access_level = 'standard'
 
 # Datasets to load inputs. If not provided, datasets will be determined
 # automatically based on the input provider implementation:
-#only_datasets = []
+#input_datasets = []
 
 # Datasets to skip:
 #skip_datasets = []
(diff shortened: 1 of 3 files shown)


vladsavelyev:
What I don't like about it is that it creates a false impression that this option is optional, e.g. the workflow picks all "datasets" by default, and this option just narrows down the list, much like the `only_samples` option does. Though we could certainly make it work this way, if Metamist would have an endpoint to tell which datasets need to be processed for e.g. a seqr-loader run. We could start working into this direction

vivbak:
Perhaps `input_datasets` is better then? Also, in this case what's the purpose of `skip_datasets`? Wouldn't you just not specify the dataset in `datasets`?

vladsavelyev:
E.g. a list of datasets would be in one config, and `skip_datasets` could be in another config.

vladsavelyev:
Let's name it `input_datasets` then.

And let's work on Metamist returning a list of datasets for seqr-loader instead of specifying it in the config

vladsavelyev (approved):

PR #94, opened 2022-10-06
Workflows: less verbose errors about missing sequence and pedigree

vladsavelyev:
No need to print every sample and cluster logs.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 4.8.7
+current_version = 4.8.8
 commit = True
 tag = False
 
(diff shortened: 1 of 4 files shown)


vivbak (commented):
Would it be helpful to still include the information about which samples have missing ped data? Maybe on one line? I think at the moment it just logs how many as opposed to which.

vladsavelyev:
@vivbak Yeah, perhaps. Though I was testing a large-cohort workflow on `prophecy-test`, and the dataset still for some reason has over a thousand samples without data, which is very annoying in logs.

vivbak (approved):
Maybe at some point we could add a 'verbose' option in the config or something to get different levels of logs

PR #96, opened 2022-10-09
Default config: remove images/hail and images/sm-api

vladsavelyev:


cpg_utils/config-template.toml
```diff
@@ -142,8 +142,9 @@ verifybamid = 'verifybamid:2.0.1'
 multiqc = 'multiqc:v1.12'
 fastqc = 'fastqc:v0.11.9_cv8'
 happy = 'hap.py:v0.3.15'
-hail = 'driver:14432044df1c5210237727cc10bee817674700c0-hail-48ff6e3a3f08a11ccb5b0ad5366ede25e43837a8'
-sm-api = 'driver:14432044df1c5210237727cc10bee817674700c0-hail-48ff6e3a3f08a11ccb5b0ad5366ede25e43837a8'
+# > skopeo copy docker://australia-southeast1-docker.pkg.dev/analysis-runner/images/<name> docker://australia-southeast1-docker.pkg.dev/cpg-common/images/<name>
+hail = 'driver:9eed154129282ac9f7adad18de6bb3f7a9f8447f-hail-7921901af5ba5fabc423f44c0d5fae46073efb9a'
+sm-api = 'driver:9eed154129282ac9f7adad18de6bb3f7a9f8447f-hail-7921901af5ba5fabc423f44c0d5fae46073efb9a'
```
lgruen: Should we really specify this in our default config? The analysis-runner already sets a the `['workflow']['driver_image']` config field -- should it also set these two images, so we don't need to update them manually?

lgruen (changes requested):

cpg_utils/config-template.toml
vladsavelyev: `driver_image` doesn't work with `image_path`, so instead of `j.image(image_path('hail'))`, we'd have to specify `j.image(get_config()['driver_image'])`, which might be confusing. But happy to change that anyway.

vladsavelyev:
Removed those `sm-api` and `hail` images, will follow up with a PR in production-pipelines

vladsavelyev:
Follow up: https://github.com/populationgenomics/production-pipelines/pull/129

lgruen (approved):

PR #100, opened 2022-10-12
Add Dockerfile and Docker CI workflow

vladsavelyev:
Currently, if we want to test recent cpg-utils changes, we have to rebuild the driver image. This change allows to run analysis-runner with `--image australia-southeast1-docker.pkg.dev/cpg-common/images/cpg_utils:4.8.12` for faster development cycle. 

It would probably be even handier to automatically build a SHA-tagged image for every pull request as well, to allow a faster development cycle in the test namespace.


.github/workflows/docker.yaml
```diff
@@ -0,0 +1,48 @@
+name: Docker
+on:
+  push:
+    branches:
+      - main
+
+env:
+  VERSION: 4.8.12
```
lgruen: Should we use the GitHub SHA as part of the tag instead? Or can we extract the version from one of the files that `bumpversion` updates?

docker/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:62e79e54d43aa1762ec68a6661872320a39517bd-hail-2ea2615a797a5aff72d20d9d12a2609342846a07
```
lgruen: Should we add a `latest` tag to the `driver` image, so we don't need to hardcode a version here (which will immediately become obsolete)?

docker/Dockerfile
```diff
@@ -0,0 +1,5 @@
...
+
+ARG VERSION=4.8.13
+
+RUN python3 -m pip install cpg_utils[workflows]==$VERSION
```
lgruen: Won't this result in a race condition with the package generation + publishing?

Should we instead build the package locally within the Docker container by copying the files from the checked out repo?

lgruen (changes requested):

.github/workflows/docker.yaml
vladsavelyev: `bump2version` updates this value directly

docker/Dockerfile
vladsavelyev: Yes, that would be amazing!

docker/Dockerfile
vladsavelyev: Good point! I'll change that

docker/Dockerfile
vladsavelyev: https://github.com/populationgenomics/analysis-runner/pull/532

docker/Dockerfile
vladsavelyev: Done

Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:62e79e54d43aa1762ec68a6661872320a39517bd-hail-2ea2615a797a5aff72d20d9d12a2609342846a07
```
lgruen:
```suggestion
FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:latest
```

Dockerfile
```diff
@@ -0,0 +1,6 @@
...
+
+ADD README.md .
+ADD cpg_utils .
+ADD setup.py .
```
lgruen: `COPY` is generally preferable to `ADD`: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy

```suggestion
COPY README.md .
COPY cpg_utils .
COPY setup.py .
```

.github/workflows/docker.yaml
lgruen: Oh, how nice, I completely missed the `[bumpversion:file:.github/workflows/docker.yaml]` part!

lgruen (approved):

PR #101, opened 2022-10-12
Add `start_query_context`

vladsavelyev:
Move a function to start Hail Query function from the large-cohort workflow codebase https://github.com/populationgenomics/large-cohort-pipeline that is in turned being moved into https://github.com/populationgenomics/production-pipelines/pull/134/files


cpg_utils/hail_batch.py
```diff
@@ -606,3 +607,45 @@ def query_command(
...
+    """
+    query_backend = query_backend or get_config().get('hail', {}).get(
+        'query_backend', 'spark'
+    )
+    if query_backend == 'spark':
+        hl.init(default_reference=genome_build())
+    elif query_backend == 'spark_local':
+        hl.init(
+            default_reference=genome_build(),
+            master='local[2]',
```
lgruen: Is `local[2]` deliberate or should this be `*`?

```suggestion
            master='local[*]',
```

lgruen (approved):

cpg_utils/hail_batch.py
vladsavelyev: `local[2]` means "run spark locally with 2 threads". `local[*]` would mean "run with all threads available" https://stackoverflow.com/questions/32356143/what-does-setmaster-local-mean-in-spark which is not a good practice for local tests, I think

vladsavelyev: I'll clarify that in the code

vladsavelyev: @lgruen added a constant for the number of threads, and a comment

PR #105, opened 2022-10-13
Fix Dockerfile

vladsavelyev:


Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM australia-southeast1-docker.pkg.dev/analysis-runner/images/driver:latest
 
 COPY README.md .
-COPY cpg_utils .
 COPY setup.py .
-RUN python3 -m pip install -e '.[workflows]'
+COPY cpg_utils cpg_utils
+RUN pip install '.[workflows]'
```
MattWellie, line 6: This should work with or without speech marks,

MattWellie (approved):

PR #115, opened 2022-10-19
When `workflow/output_version` not set, use alignments hash rather than timestamp

vladsavelyev:
Prefix stage outputs with a hash of all alignments inputs.

In other words, defaulting to _reusing_ any existing intermediates to rerunning everything. Seems like a more reasonable default for heavy workflows.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 4.9.1
+current_version = 4.9.2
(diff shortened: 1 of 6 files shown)


vivbak (approved):
Do you know what's causing the lint failure? Think a few variables just need an underscore.

vladsavelyev:
That's right! Thanks for pointing out. Fixed linting now

PR #117, opened 2022-10-26
Configuring path schemes in TOML

vladsavelyev:
Following this infra update https://github.com/populationgenomics/cpg-infrastructure/pull/20, updating the `dataset_path` function to read paths from the new generated config.

Assumes `web_url` is one of the categories, which is not implemented in cpg-infra. We can discuss if we put it elsewehere.


MattWellie:
`format()` can take multiple arguments that aren't in the format string, e.g.

```
>>> '{this}'.format(this=1, that=2)
'1'
```

so groups can substitute in whichever pre-formatting they want, and `cpg_utils` will only embed `dataset` and `main/test` if present. e.g. `gs://matts-bucket-{dataset}` will be totally valid?

vladsavelyev:
@MattWellie, sorry, not following, is there a bug somewhere? E.g. archive category is like your example: `archive = "gs://cpg-{dataset}-archive"` doesn't depend on main/test but should be valid.

cpg_utils/config-template.toml
```diff
@@ -65,59 +76,23 @@ check_intermediates = true
...
-username = 'seqr'
-# Load ElasticSearch password from a secret, unless SEQR_ES_PASSWORD is set
-password_secret_id = 'seqr-es-password'
-password_project_id = 'seqr-308602'
+# Templates to construct output storage paths. Used by `cpg_utils.hail_batch` function
+# `output_path`. The scheme below implements CPG storage policy
+# https://github.com/populationgenomics/team-docs/tree/main/storage_policies
+path_scheme = 'cpg_google_storage'
+
+[path_schemes.cpg_google_storage]
```
lgruen: Do we really need schemes here? Wouldn't it be easier to understand if these were just direct prefixes, but configured by the analysis-runner / some manual config?

E.g. why not simply `tmp = "gs://cpg-tob-wgs-main-tmp"` when you run a pipeline on the TOB-WGS dataset with a prod access level?

lgruen (changes requested):

cpg_utils/config-template.toml
vladsavelyev: @lgruen, I think I'm following, but just to be sure, can you show an example of a config like you see it, and example of `output_path` call?

MattWellie:
> @MattWellie, sorry, not following, is there a bug somewhere?

No bug! I'm really just talking to myself 😆 I was just having a play to see how tightly coupled output_path would be to the config templates

cpg_utils/config-template.toml
MattWellie: @lgruen do you mean that at runtime the config should just contain exact paths, e.g. a run using analysis-runner & the output flag `-o pinocchio` would enter `paths` into the config dictionary

```toml
[paths]
default="gs://cpg-tob-wgs-main/pinocchio"
tmp="gs://cpg-tob-wgs-main-tmp/pinocchio"
web="gs://cpg-tob-wgs-main-web/pinocchio"
...
```

these paths would still need to be configured in analysis-runner according to a set schema, even if they weren't made on the fly with each `output_path` call

lgruen: Adding suffixes at runtime sounds fine (and required for the way we use `dataset_path` at the moment), but I was just wondering whether we really need the scheme for the bucket itself. I.e. the following part should just be fixed for any given invocation, right?
```toml
[paths]
default="gs://cpg-tob-wgs-main"
tmp="gs://cpg-tob-wgs-main-tmp"
web="gs://cpg-tob-wgs-main-web"
...
```

vladsavelyev: But that would be a lot of repetition. Plus, for some workflows we don't have just one dataset, e.g. for a seqr-loader run with `dataset=seqr` we would also write to `gs://cpg-acute-care-main`, `gs://cpg-perth-neuro-main`, etc. Should that be redesigned as well?

lgruen: Great point about workflows that write to multiple dataset locations in a single run. I hadn't thought of that.

I think this means we should also replace the current `dataset_path` implementation with your more flexible approach.

vladsavelyev: > I think this means we should also replace the current dataset_path implementation with your more flexible approach.

Sorry, I'm slow today. Could you elaborate please what you mean? This PR updates `dataset_path` to use the suggested approach.

illusional:
CPG infrastructure decides this as `{path_prefix}{dataset}-{type}`

https://github.com/populationgenomics/cpg-infrastructure/blob/97d72179727f27a991a318ffae3e7b30990fe92c/stack/cpg.toml#L3

Maybe we should bind the cpg.toml onto everything else for cpg-utils to access and decide the path from here. The TOML doesn't decide the categories, but the DRIVER does, so whether the driver should bake those values in at "deploy" time and have the analysis-runner just pull from a set of defined configs, that also seems reasonable.

Maybe then you only bind azure paths on if the dataset is actually deployed to that place.

vladsavelyev:
Oh nice, I like the idea of binding to `cpg.toml`. @illusional do you think we can make sure that `get_config()` would load `cpg.toml`, passed from analysis-runner?

illusional:
I think we should move some of this heavy lifting to analysis-runner, the analylsis-runner should take the `cpg.toml` and prepare the config for that run that gives something like:

```toml
[buckets]
default = "cpg-fewgenomes-main"
tmp = "cpg-fewgenomes-main-tmp"
web = "cpg-fewgenomes-main-web"
# ...other categories I've forgotten
```

Or maybe something more specific for GCP and Azure:

```toml
[gcp.buckets]
default = "cpg-fewgenomes-main"
tmp = "cpg-fewgenomes-main-tmp"
web = "cpg-fewgenomes-main-web"
# ...other categories I've forgotten
```

Which then the `dataset_path` / `output_path` should read. Gives the advantage that:

- anyone can just prepare the final config and pass it to the pipeline, OR
- anyone that deploys our analysis-runner sees those bonuses without having to modify cpg-utils
    - AND anyone that uses our `cpg-infrastructure` gets everything for free. Not saying we should be actively promoting people using our infrastructure, but it would be a bonus if it's not much work (and actually what we submitted to ABACBS this year).


 -- OR --

Maybe the cpg-infrastructure deploy can generate a number of configs, like `fewgenomes-main.yaml` which analysis-runner can combine later. But still based on the `cpg.toml`, so above benefits are preserved - and potentially reduces the number of interaction points if we were to change it in the future :)

vladsavelyev:
Again, for workflows like seqr-loader, `output_path` and `dataset_path` could be called for multiple datasets within one invocation (e.g. `dataset=seqr` but `input_datasets=['acute-care', ...]`). So hardcoding a dataset in `"cpg-fewgenomes-main"` is not an option, unless we want to make a larger refactoring.

illusional:
Ah sorry, I missed that bit. Could it be:

```toml
[gcp.buckets]
# default is whatever dataset you ran it on 
default = "cpg-seqr-main"
tmp = "cpg-seqr-main-tmp"
web = "cpg-seqr-main-web"

# include all "dependent" datasets here?
[gcp.buckets.fewgenomes]
default = "cpg-fewgenomes-main"
tmp = "cpg-fewgenomes-main-tmp"
web = "cpg-fewgenomes-main-web"
```

vladsavelyev:
Such auto-generated config would be very long for large cohorts, so I'm not sure anyone would construct it manually, thus there wouldn't be any point of running things outside of cpg-infra. In this case we might as well make cpg-infra a package and merge it with cpg-utils, so `output_path` could generate paths dynamically :)

vladsavelyev:
Updated following the infra PR: https://github.com/populationgenomics/cpg-infrastructure/pull/20

Assumes `web_url` is one of the categories, which is not implemented in cpg-infra. We can discuss if we want to put it elsewhere.

MattWellie (approved):

illusional (approved):
I don't undersatnd why PathSchemes are being removed from here, but the other changes look good to me. Could be worth catching AttributeErrors and returnining a more friendly error, like "The category you selected was not available in this configuration".

We discussed adding a `cpg_test_dataset_path` method for subsetting. 

```python
def cpg_test_dataset_path(
    suffix: str,
    category: str | None = None,
    dataset: str | None = None,
) -> str:
    dataset = dataset or 'default'
    category = category or 'default'
    prefix = get_config()['storage'][dataset]['test'][category]
    return os.path.join(prefix, suffix)
```

vladsavelyev:
Thanks Michael! I added a flag into `dataset_path` instead, should be good?

vladsavelyev:
> I don't undersatnd why PathSchemes are being removed from here

They were only there to support generating paths for different clouds, but it's all covered in infra now

vladsavelyev:
> Could be worth catching AttributeErrors and returnining a more friendly error, like "The category you selected was not available in this configuration".

Agree, I'll add something. But globally I think we should have a more general check in the config module

vladsavelyev:
@lgruen merging is currently blocked because of your requested changes.

Screen Shot 2022-11-24 at 12 09 25

cpg_utils/hail_batch.py
```diff
@@ -78,235 +74,116 @@ def remote_tmpdir(hail_bucket: Optional[str] = None) -> str:
...
-
 def dataset_path(
     suffix: str,
-    category: Optional[str] = None,
-    dataset: Optional[str] = None,
-    access_level: Optional[str] = None,
-    path_scheme: Optional[str] = None,
+    category: str | None = None,
+    dataset: str | None = None,
+    test: bool = False,
```
lgruen: This looks like a breaking API change and should probably come with a major version bump? Have we checked whether there are many invocations that set `access_level`, which would be affected?

@illusional Didn't you prefer to have the `test` case be handled by a different method, so it's super clear that setting `test=True` breaks the portability of a pipeline and should only be used by CPG-internal pipelines?

lgruen (approved):

cpg_utils/hail_batch.py
illusional: Yes, would still prefer separate method if that’s okay.

vladsavelyev: Added a separate method

cpg_utils/hail_batch.py
```diff
@@ -78,235 +74,124 @@ def remote_tmpdir(hail_bucket: Optional[str] = None) -> str:
...
     Parameters
     ----------
     suffix : str
         A path suffix to append to the bucket.
     category : str, optional
-        A category like "upload", "tmp", "web". If omitted, defaults to the "main" and
-        "test" buckets based on the access level. See
-        https://github.com/populationgenomics/team-docs/tree/main/storage_policies
-        for a full list of categories and their use cases.
+        A category like "tmp", "web", etc., defaults to "default" if ommited.
```
lgruen:
```suggestion
        A category like "tmp", "web", etc., defaults to "default" if omited.
```

lgruen (approved):

PR #119, opened 2022-10-27
Move `workflows` library out into `cpg-workflows`

vladsavelyev:
Moving the whole `workflows` module into a new `cpg-workflows` package, following up for https://github.com/populationgenomics/production-pipelines/pull/155

Also moves out `config-template.toml` into https://github.com/populationgenomics/production-pipelines/pull/155/files#diff-f17e471fd44fe811f395fa72473d431254422e1b05906ce0d164f4bfe8250da0 as almost all stuff there was relevant to only workflows.


cpg_utils/config.py
```diff
@@ -112,8 +110,7 @@ def read_configs(config_paths: List[str]) -> frozendict:
     """
 
     config: dict = {}
-    template_path = os.path.join(os.path.dirname(__file__), 'config-template.toml')
-    for path in [template_path] + config_paths:
+    for path in config_paths:
```
lgruen, line 113: As discussed in https://centrepopgen.slack.com/archives/C018KFBCR1C/p1666918967093519, we probably want to retain the documentation purpose of `config-template.toml` somehow, but that might belong in a different repository.

lgruen (approved):

cpg_utils/config.py
vladsavelyev: @lgruen see https://github.com/populationgenomics/team-docs/pull/126 assigned to you

PR #122, opened 2022-10-31
Hail Batch command wrapper: activate gcloud creds when `define_retry_function`

vladsavelyev:
Since custom localisation uses `gsutil`, google cloud credentials must be activated.


vivbak (approved):
🧁

PR #124, opened 2022-11-10
Configs: handy functions to prepend and append; more validation

vladsavelyev:
* Add `append_config_paths` and `prepend_config_paths` in addition to `set_config_paths`
* Into `_validate_configs`, added some actual validation: parsing of all tomls and reporting errors

### DIFF
--- a/cpg_utils/config.py
+++ b/cpg_utils/config.py
@@ -19,8 +19,24 @@ def _validate_configs(config_paths: list[str]) -> None:
         raise ValueError(
             f'All config files must have ".toml" extensions, got: {config_paths}'
         )
-    if bad_files := [p for p in config_paths if not to_path(p).exists()]:
-        raise ValueError(f'Some config files do not exist: {bad_files}')
+
+    paths = [to_path(p) for p in config_paths]
+    if bad_paths := [p for p in paths if not p.exists()]:
+        raise ValueError(f'Some config files do not exist: {bad_paths}')
+
+    # Reading each file to validate syntax:
+    exception_by_path = dict()
+    for p in paths:
+        with p.open() as f:
+            try:
+                toml.loads(f.read())
+            except toml.decoder.TomlDecodeError as e:
+                exception_by_path[p] = e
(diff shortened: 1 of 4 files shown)


cpg_utils/config.py
```diff
@@ -4,29 +4,29 @@
...
+from cpg_utils import to_path
 
 # We use these globals for lazy initialization, but pylint doesn't like that.
 # pylint: disable=global-statement, invalid-name
 _config_paths = _val.split(',') if (_val := os.getenv('CPG_CONFIG_PATH')) else []
 _config: Optional[frozendict] = None  # Cached config, initialized lazily.
 
 
-def _validate_configs(config_paths: List[str]) -> None:
+def _validate_configs(config_paths: list[str]) -> None:
```
MattWellie, line 17: I know this isn't the right PR for this comment, but I'd like this validation method to do some more validation 😬 

Any thoughts on whether we should attempt to load each identified file using `toml` to check that they both exist and can be parsed correctly?

cpg_utils/config.py
```diff
@@ -45,6 +45,29 @@ def set_config_paths(config_paths: List[str]) -> None:
         _config = None  # Make sure the config gets reloaded.
 
 
+def prepend_config_paths(config_paths: list[str]) -> None:
+    """
+    Prepend the list of config paths. Equivalent to `dict.set_defaults`:
+    values in current CPG_CONFIG_PATH will have the precedence over the provided
+    `config_paths` when merging configs.
+    """
+    if _env_var := os.environ.get('CPG_CONFIG_PATH'):
```
MattWellie, line 70: love a cheeky walrus

MattWellie (approved):

cpg_utils/config.py
vladsavelyev: Yeah, good idea. Loading and parsing from remote might slow down the application a bit, but is probably worth it

vladsavelyev: I added some validation into that function

PR #127, opened 2023-01-24
Add 'get_cached_group_members' and 'check_member_in_cached_group_members'

illusional:
Functions to help facilitate checking cached group members.

Requires changes from: https://github.com/populationgenomics/analysis-runner/pull/580 to be deployed with the new infrastructure toml to be added.

### DIFF
--- a/cpg_utils/cloud.py
+++ b/cpg_utils/cloud.py
@@ -22,8 +22,9 @@ import google.api_core.exceptions
 import google.auth.transport
 from google.auth.transport import requests
 import google.oauth2
-
 from cloudpathlib import AnyPath
+
+from cpg_utils.config import get_config
 
 
 def email_from_id_token(id_token_jwt: str) -> str:
@@ -292,8 +293,6 @@ def get_cached_group_members(group) -> list[str]:
     """
     Get cached members of a group, based on the members_cache_location
     """
-    from cpg_utils.config import get_config
-
     group_name = group.split('@')[0]
     config = get_config()
 
--- a/cpg_utils/cloud.py
+++ b/cpg_utils/cloud.py
@@ -289,19 +289,22 @@ def _get_default_id_token_credentials(
     raise exceptions.DefaultCredentialsError(_HELP_MESSAGE)
 
 
-def get_cached_group_members(group) -> list[str]:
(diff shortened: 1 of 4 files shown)


cpg_utils/cloud.py
```diff
@@ -284,3 +287,25 @@ def _get_default_id_token_credentials(
...
+    """
+    group_name = group.split('@')[0]
+    config = get_config()
+
+    pathname = os.path.join(
+        config['infrastructure']['members_cache_location'], group_name + '-members.txt'
+    )
+
+    with AnyPath(pathname).open() as f:
+        return [line.strip() for line in f.readlines()]
```
lgruen: Some people also like to filter out newlines as a safeguard.

cpg_utils/cloud.py
```diff
@@ -284,3 +287,25 @@ def _get_default_id_token_credentials(
             return current_credentials
 
     raise exceptions.DefaultCredentialsError(_HELP_MESSAGE)
+
+
+def get_cached_group_members(group) -> list[str]:
```
lgruen: Could return a set here for faster subsequent checks, but I guess it doesn't matter much in our context, where we typically only check membership a single time.

lgruen (approved):

cpg_utils/cloud.py
```diff
@@ -284,3 +287,25 @@ def _get_default_id_token_credentials(
...
+
+    pathname = os.path.join(
+        config['infrastructure']['members_cache_location'], group_name + '-members.txt'
+    )
+
+    with AnyPath(pathname).open() as f:
+        return [line.strip() for line in f.readlines()]
+
+
+def check_member_in_cached_group_members(group, member) -> bool:
```
lgruen: `is_` makes the semantics a bit more clear IMO:

```suggestion
def is_member_in_cached_group(group, member) -> bool:
```

cpg_utils/cloud.py
illusional: Filtered new lines

cpg_utils/cloud.py
illusional: Returned a set per suggestion

PR #128, opened 2023-01-30
Test path generation

MattWellie:
Adds the 'test' argument to output_path and dataset_path

Where access-level is test, no impact

Where access-level != test - routes to the test storage paths in config

Adds test cases and a couple of fixtures for this and a couple of config methods

Relevant to https://github.com/populationgenomics/cpg-infrastructure/issues/39 and https://github.com/populationgenomics/analysis-runner/pull/579


requirements-test.txt
```diff
@@ -0,0 +1,2 @@
+hail
```
illusional: Do you also need a `-r requirements.txt` here too?

MattWellie:
Currently reverting changes to requirements - I split the reqs into two files so that the linting workflow wouldn't need hail to be installed, but with cache'ing that's more hassle than it's worth

illusional (commented):
LGTM, I left a comment but an update has removed the file I commented on.

illusional (approved):

PR #129, opened 2023-02-14
Make printing the configuration optional

illusional:
Clogs my logging for cpg-infrastructure, but I like using the library. I think it's reasonable to silence the logging but discourage people from doing so.


MattWellie (approved):
Seems harmless, and I like the idea of defaulting to True; I've found this pretty useful in the past, but it can push more relevant debug logging to the bottom

PR #135, opened 2023-03-20
Copy Slack utilities from cpg_workflows

lgruen:
Taken from https://github.com/populationgenomics/production-pipelines/blob/f25a8086a4f1756495faa64c1ace37df694e4851/cpg_workflows/slack.py.

This is generally useful outside of `production-pipelines`. We can point the `production-pipelines` code to this copy in a follow-up PR.

### DIFF
--- a/cpg_utils/slack.py
+++ b/cpg_utils/slack.py
@@ -8,11 +8,8 @@ Make sure `slack/channel`, `slack/token_secret_id`, and `slack/token_project_id`
 configuration values are set.
 """
 
-import os
-from textwrap import dedent
 import logging
 
-from hailtop.batch.job import Job
 import slack_sdk
 
 from cpg_utils.cloud import read_secret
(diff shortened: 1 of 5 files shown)


cpg_utils/slack.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def slack_env(j: Job):
+    """
+    Add environment variables that configure Slack reporter.
+    """
+    if not (channel := get_config().get('slack', {}).get('channel')):
+        return
+    j.env('SLACK_CHANNEL', channel)
+    j.env('SLACK_TOKEN', get_token())
```
illusional: Do we to bind the secret in here, puts it more on the public record and binds it in logs. It's not a super serious secret token, but IMO would be better to enforce getting the token from secrets each time (costs about $3 / million requests)

cpg_utils/slack.py
```diff
@@ -0,0 +1,110 @@
...
+    if not (channel := get_config()['slack'].get('channel')):
+        return
+    msg = ''
+    if data:
+        msg += '\\n'.join(f'{k}: {v}' for k, v in data.items())
+    if text:
+        msg += f'\\n{text}'
+    if not msg:
+        return
+    slack_env(j)
```
illusional: Why bind this on the environment variable when it's bound into the command explicitly below?

cpg_utils/slack.py
```diff
@@ -0,0 +1,110 @@
...
+    j.command(
+        dedent(
+            f"""
+    curl -X POST \
+    -H "Authorization: Bearer {get_token()}" \
```
illusional: Should we change this to get the secret_id and invoke gcloud directly, 

```suggestion
        TOKEN=$(gcloud secrets versions access latest --secret={SECRET_ID} --project{GCP_PROJECT}
    curl -X POST \
    -H "Authorization: Bearer $TOKEN" \
```

cpg_utils/slack.py
```diff
@@ -0,0 +1,110 @@
...
+    -H "Content-type: application/json" \
+    -d '{{"channel": "'{channel}'", "text": "{msg}"}}' \
```
illusional: Why is channel double quoted here? Another approach might be to use `json.dumps` + `shlex.quote` because the quoting will be taken care for us.

illusional (changes requested):
Requesting changes more to generate conversation around not binding the secret onto the environment variable

cpg_utils/slack.py
lgruen: I've removed everything except for the Python-based `send_message` implementation now.

illusional (approved):

PR #137, opened 2023-04-13
Get components of a GCP path

EddieLF:
Finally figured out the regex pattern to match the various components of a gcp path for almost all cpg buckets, no thanks to ChatGPT (seriously, it seems to suck at regex). 

[See here for some tests](https://regex101.com/r/r8ynHJ/1).

### DIFF
--- a/cpg_utils/cloud.py
+++ b/cpg_utils/cloud.py
@@ -325,7 +325,7 @@ def is_member_in_cached_group(
     )
 
 
-def get_path_components_from_gcp_path(path) -> dict[str, str]:
+def get_path_components_from_gcp_path(path: str) -> dict[str, str] | None:
     """
     Return the {bucket_name}, {dataset}, {bucket_type}, {subdir}, and {file} for GS only paths
     Uses regex to match the full bucket name, dataset name, bucket type (e.g. 'test', 'main-upload', 'release'),
@@ -342,18 +342,23 @@ def get_path_components_from_gcp_path(path) -> dict[str, str]:
 
     gspath_pattern = re.compile(rmatch_str)
 
-    path_components = (gspath_pattern.match(path)).groups()
+    path_match = gspath_pattern.match(path)
 
-    bucket_name = path_components[0]
-    dataset = path_components[1]
-    bucket_type = path_components[2]
-    subdir = path_components[3]
-    file = path_components[4]
+    if path_match:
+        path_components = path_match.groups()
 
-    return {
-        'bucket_name': bucket_name,
-        'dataset': dataset,
-        'bucket_type': bucket_type,
-        'subdir': subdir,
-        'file': file,
-    }
+        bucket_name = path_components[0]
+        dataset = path_components[1]
+        bucket_type = path_components[2]
+        subdir = path_components[3]
+        file = path_components[4]
+
+        return {
+            'bucket_name': bucket_name,
+            'dataset': dataset,
+            'bucket_type': bucket_type,
+            'subdir': subdir,
+            'file': file,
+        }
+
+    return None
--- a/cpg_utils/cloud.py
+++ b/cpg_utils/cloud.py
@@ -2,6 +2,7 @@
 
 import json
 import os
+import re
 import traceback
 
 from google.auth import (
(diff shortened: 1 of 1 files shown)


cpg_utils/cloud.py
```diff
@@ -322,3 +323,42 @@ def is_member_in_cached_group(
...
+    subdirectory, and the file name.
+    """
+
+    bucket_types = ['archive', 'hail', 'main', 'test', 'release']
+
+    rmatch_str = (
+        r'gs://(?P<bucket>cpg-(?P<dataset>[\w-]+)-(?P<bucket_type>['
+        + '|'.join(s for s in bucket_types)
+        + r']+[-\w]*))/(?P<suffix>.+/)?(?P<file>.*)$'
+    )
```
MattWellie: this is cool, but unless it's parametrised from config you may as well just make the fixed string here?

cpg_utils/cloud.py
```diff
@@ -322,3 +323,42 @@ def is_member_in_cached_group(
...
+
+        return {
+            'bucket_name': bucket_name,
+            'dataset': dataset,
+            'bucket_type': bucket_type,
+            'subdir': subdir,
+            'file': file,
+        }
+
+    return None
```
MattWellie: I'd be happy to either return contents or fail here. It's a very specific method, and being able to return `None` in what is probably an error state just makes the output a bit harder to predict elsewhere

```suggestion

    if path_match := gspath_pattern.match(path):
        path_components = path_match.groups()

        bucket_name = path_components[0]
        dataset = path_components[1]
        bucket_type = path_components[2]
        subdir = path_components[3]
        file = path_components[4]

        return {
            'bucket_name': bucket_name,
            'dataset': dataset,
            'bucket_type': bucket_type,
            'subdir': subdir,
            'file': file,
        }

    raise ValueError('The input String did not match a valid GCP path')

```

cpg_utils/cloud.py
```diff
@@ -322,3 +323,42 @@ def is_member_in_cached_group(
...
+        subdir = path_components[3]
+        file = path_components[4]
+
+        return {
+            'bucket_name': bucket_name,
+            'dataset': dataset,
+            'bucket_type': bucket_type,
+            'subdir': subdir,
+            'file': file,
+        }
```
MattWellie: I just discovered this, neat shortcut 

```suggestion
        return path_match.groupdict()
```

MattWellie (changes requested):
I think a failure to match should be an error, not a None. The other suggestions are syntax changes but the functionality is the same

MattWellie:
Also all the suggestions I added overlap in a really stupid way
...

```python
def get_path_components_from_gcp_path(path: str) -> dict[str, str]:
    """
    Return the {bucket_name}, {dataset}, {bucket_type}, {subdir}, and {file} for GS only paths
    Uses regex to match the full bucket name, dataset name, bucket type (e.g. 'test', 'main-upload', 'release'),
    subdirectory, and the file name.
    """

    bucket_types = ['archive', 'hail', 'main', 'test', 'release']

    # compile pattern matching all CPG bucket formats
    gspath_pattern = re.compile(
        r'gs://(?P<bucket>cpg-(?P<dataset>[\w-]+)-(?P<bucket_type>['
        + '|'.join(s for s in bucket_types)
        + r']+[-\w]*))/(?P<suffix>.+/)?(?P<file>.*)$'
    )

    # if a match succeeds, return the key: value dictionary
    if path_match := gspath_pattern.match(path):
        return path_match.groupdict()

    # raise an error if the input String was not a valid CPG bucket path
    raise ValueError('The input String did not match a valid GCP path')
```

cpg_utils/cloud.py
EddieLF: Not really sure what you mean by this, are you saying to scrap the bucket_types list and add them into the string?

cpg_utils/cloud.py
EddieLF: Now this I love

cpg_utils/cloud.py
MattWellie: Pretty much. I actually like keeping this in, as it shows the syntax if we want to entertain a wider range of bucket categories later.

lgruen:
Have you tried https://regex.ai/?

EddieLF:
> Have you tried https://regex.ai/?

You keep giving me amazing tools to bookmark @lgruen! This is another awesome one, alongside phind.com

MattWellie (approved):

PR #141, opened 2023-10-02
Pinning boto3/botocore versions to prevent downloading all the boto3 …

milo-hyben:
…versions when installing cloudpathlib.


illusional (approved):
Can you add a bumpversion commit please 🙏

PR #144, opened 2023-10-17
Restrict docker push to main only

illusional:
I don't think this image is used, but rather than delete, I'll just patch the non-main workaround for this and the package command


MattWellie (approved):
Also unsure if this has been used in production as an image, but happy with the changes

PR #145, opened 2023-10-23
Migrate get_batch wrapper to cpg-utils

MattWellie:
Steals the get_batch wrapper structure from cpg_workflows (library within production-pipelines)

Placement of this content in cpg-utils will remove a direct dependency between ad-hoc analysis workflows and the core production-pipelines library

### DIFF
--- a/cpg_utils/hail_batch.py
+++ b/cpg_utils/hail_batch.py
@@ -6,11 +6,7 @@ import logging
 import os
 import tempfile
 import textwrap
-import uuid
-
-from typing import List, Literal, Optional, Union
-
-import toml
+from typing import Dict, List, Literal, Optional, Union
 
 import hail as hl
 import hailtop.batch as hb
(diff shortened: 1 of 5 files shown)


cpg_utils/hail_batch.py
```diff
@@ -28,6 +35,228 @@
...
+        if not get_config()['hail'].get('dry_run') and not isinstance(
+            self._backend, hb.LocalBackend
+        ):
+            self._copy_configs_to_remote()
+
+    def _copy_configs_to_remote(self):
+        """If configs are local files, copy them to remote"""
+        remote_dir = to_path(self._backend.remote_tmpdir) / 'config'
+        config_path = remote_dir / (str(uuid.uuid4()) + '.toml')
+        with config_path.open('w') as f:
```
illusional: Do we need this, or want to encourage this?

illusional (approved):
This is awesome! Can we by default ensure the created batch has the ar-guid as a batch attribute?

cpg_utils/hail_batch.py
MattWellie: I wasn't sure about this - I'd be happy to remove if I knew it wouldn't break anything.

My thinking on this was that analysis-runner already handled generating the run config and posting to GCP, this would only be useful for jobs where we want to use `get_config` but don't have the libraries required to read from GPC?

PR #146, opened 2023-11-14
Don't print config content by default

MattWellie:
This is very petty...

The config module is currently set up to print the config contents by default each time a new config is loaded. In the pipeline startup the config content is repeated 3 times resulting in a huge wall of redundant text. This also clouds the logging for failed tests etc. and is generally not useful.

The first get_config invocation in the pipeline is easy to find, but I haven't located the subsequent 2 calls that trigger re-printing of the same content...

Instead of mucking about in the prod-pipes codebase to track down the exact calls, how do you feel about treating this at source - alter the default behaviour not to print to logging, then set `get_config(True)` as a deliberate action (e.g. the first time the config is loaded, then never again)

Proposal generally approved on Slack: https://centrepopgen.slack.com/archives/C018KFBCR1C/p1700002730474369

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 4.18.0
+current_version = 4.18.1
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -5,7 +5,7 @@ on:
       - main
 
 env:
-  VERSION: 4.18.0
+  VERSION: 4.18.1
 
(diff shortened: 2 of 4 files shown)


illusional (commented):
I'm in favour of this, but can we consult before we make this change. Can you post in team data, gather some feedback and we can assess based on that?

MattWellie:
> Can you post in team data, gather some feedback and we can assess based on that?

Aye, I'll post tomorrow. Figured you fine folk would be a good litmus test 💪

cassimons (approved):

PR #147, opened 2023-11-22
Add reset_batch method

illusional:
Add a mechanism to reset the global batch reference, potentially useful for tests.


MattWellie:
Ahhhhh I thought I tried this but it didn't work. Maybe I tried importing the global variable and resetting it in the importing code instead of at source. Will check this morning

MattWellie (approved):
added a test case, checked locally that it works

PR #148, opened 2023-12-03
Driver as default bash image

MattWellie:
We don't set a default image for BashJob (and we never really did, though some jobs did use `hb.Batch(default_image=XXX)` when creating a batch manually)

This changes the batch creation wrapper to set the driver image as the bash job default if not otherwise specified, which would otherwise default to a naked ubuntu:22

### DIFF
--- a/cpg_utils/hail_batch.py
+++ b/cpg_utils/hail_batch.py
@@ -83,7 +83,8 @@ def get_batch(
             cancel_after_n_failures=get_config()['hail'].get('cancel_after_n_failures'),
             default_timeout=get_config()['hail'].get('default_timeout'),
             default_memory=get_config()['hail'].get('default_memory'),
-            default_image=default_bash_image or get_config()['workflow']['driver_image'],
+            default_image=default_bash_image
+            or get_config()['workflow']['driver_image'],
             default_python_image=default_python_image
             or get_config()['workflow']['driver_image'],
             attributes=attributes,
--- a/cpg_utils/hail_batch.py
+++ b/cpg_utils/hail_batch.py
@@ -41,8 +41,8 @@ def reset_batch():
 def get_batch(
     name: str | None = None,
     *,
+    default_image: str | None = None,
     default_python_image: str | None = None,
-    default_bash_image: str | None = None,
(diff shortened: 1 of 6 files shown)


illusional:
I'm split between whether this is a good idea or not, do we want people to explicitly choose which container they should use? It's easy, but it's potentially promoting idle behaviour.

Maybe we could discuss this at data office hours?

MattWellie:
> do we want people to explicitly choose which container they should use?

I think yes... but this is just the default which in most cases will probably be overridden. There are a lot of situations where we're creating images and setting the image as the driver image anyway, in which case this is a simplification. 

We've loaded a ton of libraries into the driver, so it makes sense as a base for a lot of jobs, but yeah... this potentially paves the way for a lot of 'it just works... and I don't know why'

cpg_utils/hail_batch.py
```diff
@@ -81,6 +83,8 @@ def get_batch(
             cancel_after_n_failures=get_config()['hail'].get('cancel_after_n_failures'),
             default_timeout=get_config()['hail'].get('default_timeout'),
             default_memory=get_config()['hail'].get('default_memory'),
+            default_image=default_bash_image
+            or get_config()['workflow']['driver_image'],
```
illusional:
```suggestion
            default_image=default_bash_image,
```

illusional (approved):
I've approved the change in behaviour because I think it's blocking, but we should retrospectively modify this behaviour based on the outcome of tomorrow's or the following week's conversation. 

Can you add a in-line comment to note this?

cpg_utils/hail_batch.py
MattWellie: I'm fine to hold off on this for now, and settle after discussion tomorrow. @EddieLF has adapted the failing script already so nothing urgent.

Unsure about this suggestion enabling None as a pass-thru, that way we'd still end up back where we started with the default image being a naked Ubuntu, which is almost never going to have the required libraries installed.

illusional: I didn't initially see the kwargs at the bottom, otherwise It just makes it clearer you can pass the default_image. In this case, let's pause.

PR #149, opened 2023-12-12
Write cloud config and overwrite env variable

MattWellie:
See https://centrepopgen.slack.com/archives/C030X7WGFCL/p1702345700581549

When the Hail Batch wrapper was migrated from production-pipelines to cpg-utils we removed one function call that seemed unnecessary. The final action when creating a batch was to make a config file and write to a cloud location.

It turns out this is pretty important - Analysis-runner generates one combined config from all the CLI file paths. When cpg-workflows/main.py runs it then combines that AR config with the default config and workflow-specific config, adding all 3 paths to the CPG_CONFIG_PATH env variable (based on local files within production-pipelines).

This missing step (reintroduced by this PR) combines all those 3 files into one config, writes it to a cloud path, then overwrites the env variable with a single path. This is then passed on to all child jobs in the batch.

End product - all child jobs get one config file path they can read remotely, which is a combination of the Driver job config and all workflow-specific configs.

### DIFF
--- a/cpg_utils/hail_batch.py
+++ b/cpg_utils/hail_batch.py
@@ -6,8 +6,10 @@ import logging
 import os
 import tempfile
 import textwrap
+import uuid
 from typing import Dict, List, Literal, Optional, Union
 
+import toml
 import hail as hl
 import hailtop.batch as hb
 from hail.utils.java import Env
@@ -18,6 +20,7 @@ from cpg_utils.config import (
     ConfigError,
     get_config,
     retrieve,
+    set_config_paths,
     try_get_ar_guid,
 )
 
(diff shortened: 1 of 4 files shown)


jmarshall (approved):
Looks good — just like the one in _cpg_workflows/batch.py_ in fact :smile:

cpg_utils/hail_batch.py
```diff
@@ -6,8 +6,10 @@
 import os
 import tempfile
 import textwrap
+import uuid
```
jmarshall, line 9: Thought that crazy linter would want this after `from typing …` but looks like it's me who needs a refresher on which way around `import AB` and `from AA import X` are supposed to go…

jmarshall:
:+1: for “why do we need this” comments

cpg_utils/hail_batch.py
MattWellie: Yeah, afaik it's python builtin `import X`, `from X`, then the same for external libs, then for local/private libs. I'm still not sure if capitalised should come before lower case, I just tweak it until the linter gives me the green light 😆

illusional:
Hmm, I didn’t love the idea of rewriting the config as it does make tracking it back through the analysis runner more difficult. 

Can we make it not rewrite if it’s not actually combining this

Happy with this change to resolve the problem, but we should flag this as debt to pay back.

PR #152, opened 2024-03-13
Refactor repository + add tests

illusional:
- Rehome new functions (will provide a better diff at the end)
- Move some functions from analysis-runner to here (cromwell, git, dataproc)
- Add tests
- Upgrade linting
- Decreased reliance on analysis-runner project
    - https://github.com/populationgenomics/cpg-infrastructure/pull/212

I intend to create a docs site from team-docs, that documents the cpg-utils library.


cpg_utils/cloudpath_hail_az.py
```diff
@@ -72,7 +72,9 @@ def __init__(
 
     @classmethod
     def is_valid_cloudpath(
-        cls, path: Union[str, CloudPath], raise_on_error=False
+        cls,
+        path: Union[str, CloudPath],
```
nevoodoo, line 76: Should we change this to `path: str | CloudPath`?

nevoodoo: Should we change this to `path: str | CloudPath`? And we can replace the use of `Optional` in the `init` method of `HailAzureBlobPath` too :)

cpg_utils/git.py
```diff
@@ -2,13 +2,12 @@
...
 """
 
-
-from typing import List
-
 import os
 import re
 import subprocess
-from shlex import quote
+from typing import List
```
nevoodoo: Maybe we replace the use of this with `list[str]` if the Python version for this repo supports it :D

cpg_utils/cromwell_model.py
```diff
@@ -0,0 +1,438 @@
...
+        id=None,
+        inputs=None,
+        labels=None,
+        submission=None,
+        status=None,
+        end=None,
+        start=None,
+        **kwargs,
+    ):
+        self.workflowName = workflowName
```
nevoodoo, line 87: Not a review comment per se but is camel case okay to use? I just get thrown off by it in python files but as long as everyone is okay with its usage, I'm okay :)

cpg_utils/hail_batch.py
```diff
@@ -824,21 +706,87 @@ def start_query_context(
...
+            ),
         )
 
 
-def cpg_namespace(access_level) -> str:
-    """
-    Get storage namespace from the access level.
-    """
-    return 'test' if access_level == 'test' else 'main'
+def run_batch_job_and_print_url(batch: Batch, wait: bool, environment: str):
```
nevoodoo: Should this (and the methods below it) have return types?

nevoodoo (commented):
Just added some comments for things I could easily pick up :) Don't think I see anything else from a code POV!

nevoodoo (approved):

cpg_utils/constants.py
```diff
@@ -0,0 +1,37 @@
...
+DEFAULT_CROMWELL_AUDIENCE = (
+    '717631777761-ec4u8pffntsekut9kef58hts126v7usl.apps.googleusercontent.com'
+)
+
+
+CROMWELL_AUDIENCE = os.getenv('CROMWELL_AUDIENCE', DEFAULT_CROMWELL_AUDIENCE)
+CROMWELL_URL = os.getenv('CROMWELL_URL', DEFAULT_CROMWELL_URL)
+
+
+class AnsiiColors:
```
jmarshall, line 24: Where is this used, and is this an opportunity to fix the spelling to `AnsiColors`?

(Or even `AnsiColours`! 😝 )

PR #153, opened 2024-03-13
Quote here-document delimiters to avoid expansion

jmarshall:
Without quotes, `{script_contents}` and `{python_code}` are subject to `$var`, \`command\`, and `$((expr))` expansion by the shell.

Context: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1710309190240709?thread_ts=1710299130.937109&cid=C030X7WGFCL

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 4.19.0
+current_version = 4.19.1
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -5,7 +5,7 @@ on:
       - main
 
 env:
-  VERSION: 4.19.0
+  VERSION: 4.19.1
 
 jobs:
   docker:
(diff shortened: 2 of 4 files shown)


MattWellie:
Looks like this will solve the interpretation of comments as bash variables - I hate this method design, and would be in favour of ditching the "cat all lines to a file then run that file" approach. Adding a separate script to replace the query_command wrappers makes sense to me, each of these jobs then just runs a script which is predictable and in version control instead of needing a runtime module inspection. Probably no the time or place, but I'll make an issue

MattWellie (approved):
I've tried this locally, it's a good band-aid but the process is bad

jmarshall:
Yes :smile:  We've put in a hail feature request hail-is/hail#14354 to have a facility to get scripts uploaded to the worker, and having the hard part done for us leaves it easy enough to add a little UI friendliness on top in a wrapper if needed. So in the fullness of time this will get better; in the meantime improving the transport via here-documents is a win…

jmarshall:
Hmmm… maybe this deserves a little bumpversioning…

PR #154, opened 2024-03-14
Convert workflow to use workload identity federation and update actions versions

jmarshall:
My recent PR's [Docker action run](https://github.com/populationgenomics/cpg-utils/actions/runs/8274277095) failed with

```
google-github-actions/setup-gcloud failed with: failed to execute command `gcloud --quiet auth activate-service-account gh-images-deployer@cpg-common.iam.gserviceaccount.com --key-file -`:
ERROR: (gcloud.auth.activate-service-account) There was a problem refreshing your current auth tokens:
('invalid_grant: Invalid JWT Signature.', ***'error': 'invalid_grant', 'error_description': 'Invalid JWT Signature.'***)
```

Possibly `secrets.GH_IMAGES_DEPLOYER_JSON` has expired?

This PR converts the workflow to use workload identity federation instead. I've copied `workload_identity_provider` and `service_account` from elsewhere — @illusional perhaps we need to generate new ones for this workflow in this different repository…?

While we're here, update all the other pre-canned actions used to their latest (non-deprecated) node20-based versions.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -7,6 +7,10 @@ on:
 env:
   VERSION: 4.19.1
 
+permissions:
+  contents: read
+  id-token: write
+
 jobs:
   docker:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


jmarshall:
Oh, I see it's been failing for a few months: https://github.com/populationgenomics/cpg-utils/actions/workflows/docker.yaml

.github/workflows/docker.yaml
```diff
@@ -19,15 +23,16 @@ jobs:
...
+    - uses: actions/checkout@v4
 
     - name: gcloud auth
-      uses: 'google-github-actions/auth@v0'
+      uses: google-github-actions/auth@v2
       with:
-        credentials_json: ${{ secrets.GH_IMAGES_DEPLOYER_JSON }}
-  
+        workload_identity_provider: "projects/1051897107465/locations/global/workloadIdentityPools/github-pool/providers/github-provider"
+        service_account: "gh-images-deployer@cpg-common.iam.gserviceaccount.com"
```
jmarshall, line 32: Possibly `project_id: cpg-common` should be up here instead / as well.

illusional (approved):
Looks good to me! I've added the workload identity user to the gh-images-deploy. Fun that you can do this on the UI through the IAM and admin -> Service Accounts page.

I'm pretty confident I removed the old credentials, partially to track where the image was being used.

PR #156, opened 2024-04-14
Don't include dependencies from requirements.txt

illusional:
- requirements.txt is not packaged: https://github.com/populationgenomics/cpg-utils/actions/runs/8679213605/job/23797477411
- Rename ANSII to ANSI
- Move Cromwell test from analysis-runner to here


jmarshall:
Interesting. I was going to say in pysam we have a _MANIFEST.in_ file that controls adding a bunch of extra files to the sdist. Hence cpg-utils could do that (or perhaps use `setup(…, package_data=…)` similarly) to ensure _requirements.txt_ is present.

And then I skimmed [examples-of-why-parsing-requirements-txt-files-to-feed-project-metadata-is-a-bad-idea](https://discuss.python.org/t/examples-of-why-parsing-requirements-txt-files-to-feed-project-metadata-is-a-bad-idea/24828/14) and the blog posting linked from there, which suggests reading _requirements.txt_ in _setup.py_ may be an anti-pattern… 🤔

illusional:
Yeah I read that article too, I put the `.` in the requirements.txt to practically install any requirements (backwards though).

nevoodoo (approved):

MattWellie:
I've always gotten around the installation with manifest packaging of a requirements file... I understand that the consensus is that it's a bad idea, but with no forked repositories or private packages, is there a difference between the two approaches here?

illusional:
Not really @MattWellie, except I didn't have to test as much by putting it in the setup.py because I know it will work (and it mirrors what was previously here).

Issue #157, opened 2024-04-16
Dataproc startup borked - trailing slash


MattWellie:
https://github.com/populationgenomics/cpg-utils/blob/b42013b75e6812335a11ba1201cd2b9f495a4fa9/cpg_utils/dataproc.py#L284-L297

see https://batch.hail.populationgenomics.org.au/batches/442999/jobs/90

This join onto an empty String results in a trailing slash

```
>>> dataset_path('', 'tmp').split('://')[1]
'cpg-PROJECT-main-tmp/'
```

This is not tolerated during dataproc startup:

```
ERROR: (gcloud.dataproc.clusters.create) INVALID_ARGUMENT: Storage Bucket name 'cpg-PROJECT-main-tmp/' must match pattern '[a-z0-9\-_.]{3,222}'
```

## proposal

remove the dataset-path call with

```
get_config()['storage']['default']['tmp'].removeprefix('gs://')
```

This would also solve the production-pipelines test use-case - config files are generated during tests, having tmp-dir test paths embedded as storage options (these do not start with `gs://`). I patched this [here](https://github.com/populationgenomics/production-pipelines/pull/690/commits/3c9188e9d8095632b113fbd3fa635c6d90cdf409) to get the test suite running, but a change in this lib would make that unnecessary.

EddieLF:
We could use the new `config_retrieve` function like `config_retrieve(['storage', 'default', 'tmp'])` to get the bucket name

PR #158, opened 2024-04-16
Fix dataproc startup (temp folder)

MattWellie:
As described in #157

### DIFF
--- a/cpg_utils/dataproc.py
+++ b/cpg_utils/dataproc.py
@@ -10,6 +10,7 @@ import hailtop.batch as hb
 
 from cpg_utils.config import (
     AR_GUID_NAME,
+    config_retrieve,
     dataset_path,
     get_access_level,
     get_gcloud_set_project,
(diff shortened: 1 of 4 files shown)


cpg_utils/dataproc.py
```diff
@@ -10,6 +10,7 @@
 
 from cpg_utils.config import (
     AR_GUID_NAME,
+    config_retrieve,
     dataset_path,
```
EddieLF:
```suggestion
```

EddieLF:
Will there need to be some handling for test runs? 
The test temp bucket is at ['storage', 'default', 'test', 'tmp'], might we need something like:

```python
if get_access_level() == 'test':
  tmp_bucket_config_list = ['storage', 'default', 'test', 'tmp']
else:
  tmp_bucket_config_list = ['storage', 'default', 'tmp']

tmp_bucket = config_retrieve(tmp_bucket_config_list)
```

MattWellie:
> Will there need to be some handling for test runs?

I believe this is handled intelligently in config generation - test runs don't have main storage locations, so storage/default/tmp in a test run will point to a test path. In a standard/full permission run it will point to a main path

EddieLF (approved):

PR #160, opened 2024-04-16
Allow config_retrieve to supply default if no config is available

illusional:
Action that is failing as the behaviour changed: https://github.com/populationgenomics/metamist/actions/runs/8697673597/job/23870558763


vivbak (approved):
Thanks!

PR #161, opened 2024-04-17
Environment Lock

nevoodoo:
## Description
In a bid to introduce better security permissions, we have replaced the use of SA keys to workload identity providers. This change will also implement the use of environments to ensure deployments are going through the right channels.

## Changes
- [x]  Added `environment` attribute to the `docker` workflow


illusional (approved):

illusional:
Can you apply the same treatment to the package workflow, and then move the pypi API token to the production environment (maybe in a follow-up PR).

PR #164, opened 2024-04-19
Use basicConfig setup for logging.info messages

MattWellie:
Follow-up fix on top of https://github.com/populationgenomics/cpg-utils/pull/162

See https://batch.hail.populationgenomics.org.au/batches/443997/jobs/2

By importing Logging, the native logging level is `Warning`. The `info` level messages aren't printed, this includes printing the workflow ID and regular status updates. The first update in a Cromwell-watcher job is during failure (warning-level logging/print statements, everything up to that is logging-level, so not printed to logs)

This change runs a basicConfig to change the logging level to info, which will cause this information to be presented in Hail Batch logs.

During the import from analysis-runner to cpg-utils we went from [this](https://github.com/populationgenomics/analysis-runner/blob/47044d5036f3930c6796a602ae314fa842298dbc/analysis_runner/util.py#L16-L21):

```
logger = logging.getLogger('analysis_runner')
logger.addHandler(logging.StreamHandler())
logger.setLevel(logging.INFO)
# Also update the default severity level for modules that don't use the
# 'analysis-runner' logger.
logging.getLogger().setLevel(logging.INFO)
```

to the simplified `logger =  logging.getLogger(...)` which is how we lost this logging setup.

### DIFF
--- a/cpg_utils/cromwell.py
+++ b/cpg_utils/cromwell.py
@@ -5,7 +5,6 @@ jobs from within Hail batch.
 """
 
 import json
-import logging
 import os
 import subprocess
 from shlex import quote
@@ -34,8 +33,6 @@ from cpg_utils.git import (
     get_repo_name_from_remote,
 )
 from cpg_utils.hail_batch import prepare_git_job, query_command
-
-logger = logging.getLogger(__name__)
 
 
 class CromwellOutputType:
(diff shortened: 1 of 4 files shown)


cpg_utils/cromwell.py
```diff
@@ -386,6 +386,9 @@ def watch_workflow(  # noqa: C901
     )
     from cpg_utils.cromwell_model import WorkflowMetadataModel
 
+    # ensure logging info statements are actually printed
+    logging.basicConfig(level=logging.INFO)
```
illusional: This will apply to all loggers, could we potentially revert your previous change, to create an info logger just for this file?

illusional (changes requested):

cpg_utils/cromwell.py
MattWellie: It does, but this method is executed as a distinct script in a separate batch job so there is only one logger. I'll add the same logging setup as before, but `__file__` is not a meaningful name here, as we're creating `script.py`

illusional (approved):

illusional:
@MattWellie you'll just need a bumpversion to release this

cpg_utils/cromwell.py
MattWellie: I've re-added the logger instance generation, using the name `cromwell_watcher`

PR #165, opened 2024-04-24
Abort overlong jobs

MattWellie:
After discussion of billing, rogue samples taking 20x longer than they should (and rarely completing) are costing 20x more. Seems obvious, but it wasn't certain if the time taken was idle compute/getting lost in an event queue. 

Specific use-case here is GATK-SV, where running Manta on some individual CRAMs has the same compute cost as processing 150 successful samples, and no results are ever generated.

This PR proposes a kill-switch, so we can set each type of job to have an expected life expectancy, and we can abort jobs which appear to be in an endless loop without having to manually monitor. 

All decisions about the appropriate time permitted will be per-job, and decided in the pipeline. This change only permits an integer to be passed through when starting a Cromwell task which will detect and abort overlong jobs

### DIFF
--- a/cpg_utils/cromwell.py
+++ b/cpg_utils/cromwell.py
@@ -287,6 +287,7 @@ def run_cromwell_workflow_from_repo_and_get_outputs(
     copy_outputs_to_gcp: bool = True,
     min_watch_poll_interval: int = 5,
     max_watch_poll_interval: int = 60,
+    time_limit_seconds: int | None = None,
 ) -> tuple[Job, dict[str, Resource | list[Resource]]]:
     """
     This function needs to know the structure of the outputs you
@@ -309,6 +310,9 @@ def run_cromwell_workflow_from_repo_and_get_outputs(
 
     Optionally override min/max poll interval for the watch job.
     This alters how often the Watch job pings Cromwell for Status updates
+
+    time_limit_seconds is optional, and will cause the workflow to be aborted if
+    the time limit is exceeded
     """
 
     _driver_image = driver_image or get_driver_image()
@@ -346,6 +350,7 @@ def run_cromwell_workflow_from_repo_and_get_outputs(
         driver_image=_driver_image,
         max_poll_interval=max_watch_poll_interval,
         min_poll_interval=min_watch_poll_interval,
+        time_limit_seconds=time_limit_seconds,
     )
 
     return submit_job, outputs_dict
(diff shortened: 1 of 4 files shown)


cpg_utils/cromwell.py
```diff
@@ -429,21 +445,38 @@ def _get_wait_interval(
...
+    cumulative_wait: int = 0
 
     while True:
+        # kill the workflow if the maximum wait has elapsed
+        if time_limit_seconds and cumulative_wait > time_limit_seconds:
+            # time to die, Mr. Cromwell
+            auth_header = {'Authorization': f'Bearer {_get_cromwell_oauth_token()}'}
+            r = requests.post(abort_url, headers=auth_header, timeout=10)
+            if not r.ok:
+                logger.info(f"You couldn't even die right, {workflow_id}")
```
MattWellie: I don't expect this to reach main, but I'm really not sure how to handle unkillable eternal jobs 💸

illusional: Should probably wrap this in some retryability

cpg_utils/cromwell.py
```diff
@@ -429,21 +445,38 @@ def _get_wait_interval(
     status_reported = False
     subprocess.check_output(GCLOUD_ACTIVATE_AUTH_BASE)  # noqa: S603
     cromwell_workflow_root = f'{CROMWELL_URL}/api/workflows/v1/{workflow_id}'
+    abort_url = f'{cromwell_workflow_root}/abort'
     metadata_url = f'{cromwell_workflow_root}/metadata'
     outputs_url = f'{cromwell_workflow_root}/outputs'
     status_url = f'{cromwell_workflow_root}/status'
     _remaining_exceptions = max_sequential_exception_count
     start = datetime.now()
+    cumulative_wait: int = 0
```
illusional: Use a DateTime start here, and check time elapsed so you're not waiting for long requests.

cpg_utils/cromwell.py
MattWellie: A failure here will fail over to the next loop, where the abort will be retried. That will indirectly provide `max_sequential_exception_count` abort attempts before the watcher errors out. Happy to leave handling there, or hang here and retry the abort in a separate loop?

illusional: Could you list the exception / status code in this warning?

cpg_utils/cromwell.py
MattWellie: Screenshot 2024-04-24 at 5 26 09 PM
So rad that PR comments are now showing up in PyCharm

MattWellie:
An alternative to this would be setting a global pipeline timeout at the Cromwell level, though that would affect all users, and isn't runtime configurable (e.g. we have jobs that naturally take over 24 hours, but variant calling in particular shouldn't take that long)
https://github.com/broadinstitute/cromwell/pull/5273/files

illusional (approved):

PR #166, opened 2024-05-06
Accept Path as well as str in read_input() and read_input_group()

jmarshall:
populationgenomics/production-pipelines#727 was yet another example of us forgetting to call `str()` on a CloudPath before passing it to `read_input()`. I do this regularly too, but usually catch it (with great tedium) in local testing.

I'm not sure there's any good reason why we do this to ourselves! Seems like there'd be no conceptual reason why these methods **shouldn't** take a path-like equivalently to a `str`, so this PR suggests overriding these methods in our `hb.Batch` subclass so that they will accept either `str` (as per the superclass) or additionally `CloudPath`/`pathlib.Path`.

(It would be less useful to push this upstream to mainstream Hail, as it doesn't already use cloudpathlib so they would probably prefer to only support `str | pathlib.Path`. But I suppose an alternative to this subclass addition would be to get that much merged upstream and then carry `… | CloudPath` in our local Hail patches…)

### DIFF
--- a/cpg_utils/hail_batch.py
+++ b/cpg_utils/hail_batch.py
@@ -155,6 +155,17 @@ class Batch(hb.Batch):
             toml.dump(dict(get_config()), f)
         set_config_paths([str(config_path)])
 
+    def read_input(self, path: str | Path) -> hb.resource.InputResourceFile:
+        """As per Hail's read_input() but allows either str or Path."""
+        assert isinstance(path, str | Path)
+        return super().read_input(str(path))
+
+    def read_input_group(self, **kwargs: str | Path) -> hb.resource.ResourceGroup:
+        """As per Hail's read_input_groups() but allows either str or Path."""
+        for path in kwargs.values():
+            assert isinstance(path, str | Path)
+        return super().read_input_group(**{k: str(v) for k, v in kwargs.items()})
+
     def _process_job_attributes(
         self,
         name: str | None = None,


cpg_utils/hail_batch.py
```diff
@@ -155,6 +155,17 @@ def _copy_configs_to_remote(self) -> None:
             toml.dump(dict(get_config()), f)
         set_config_paths([str(config_path)])
 
+    def read_input(self, path: str | Path) -> hb.resource.InputResourceFile:
+        """As per Hail's read_input() but allows either str or Path."""
+        assert isinstance(path, str | Path)
```
jmarshall, line 160: Or perhaps drop the asserts and depend on type-checking to ensure this.

illusional: In favour of this change, I've always been a bit weary of asserts as they can be optimised out: https://stackoverflow.com/questions/1273211/disable-assertions-in-python
Not that we actively do this though, and I guess it's then a no-op change from before.

jmarshall:
Actually maybe the upstream Hail change that would render this overriding unnecessary would be

```diff
--- a/hail/python/hailtop/batch/batch.py
+++ b/hail/python/hailtop/batch/batch.py
@@ -466,7 +466,7 @@ class Batch:
         self._resource_map[jrf._uid] = jrf  # pylint: disable=no-member
         return jrf
 
-    def read_input(self, path: str) -> _resource.InputResourceFile:
+    def read_input(self, path: str | os.PathLike) -> _resource.InputResourceFile:
         """
         Create a new input resource file object representing a single file.
 
@@ -495,7 +495,7 @@ class Batch:
         irf = self._new_input_resource_file(path)
         return irf
 
-    def read_input_group(self, **kwargs: str) -> _resource.ResourceGroup:
+    def read_input_group(self, **kwargs: str | os.PathLike) -> _resource.ResourceGroup:
         """Create a new resource group representing a mapping of identifier to
         input resource files.
```

along with a suitable (check and) conversion to `str`, perhaps in `_new_input_resource_file()`.

Both `pathlib.Path` and `CloudPath` are path-like as measured by `os.PathLike`, so this would do the trick for cloudpathlib without Hail needing to mention it directly.

illusional (approved):
Would need a bumpversion commit to release this, I'm broadly supportive of this change, especially in the midterm while the linters aren't completely set-up.

jmarshall:
The upstream version of this has now been merged upstream: see hail-is/hail#14544. I would be tempted to cherry-pick that into our current Hail uplift so that we don't have to apply this version to our subclass after all.

illusional:
@jmarshall +1 to cherry picking this commit (or merging up to it!)

jmarshall:
The upstream version of this appeared in 0.2.131 so it has been available in our installation for some time. So this PR is no longer necessary.

PR #167, opened 2024-05-09
Add Hail Default version for Dataproc

MattWellie:
It would be unusual for us to provide a non-default hail version for a dataproc cluster, so should we set a default here?

n.b. our prod-pipes code for submitting jobs to an existing cluster doesn't specify this attribute. I don't think we ever use that code path, and due to [this](https://github.com/populationgenomics/production-pipelines/pull/728) we're not failing a mypy inspection on cpg-utils methods.

e.g. [here](https://github.com/populationgenomics/production-pipelines/blob/b8b10456ca0b74d1c60bd8f815684073c7fe6132/cpg_workflows/large_cohort/dataproc_utils.py#L71-L79)

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 5.0.8
+current_version = 5.0.9
 commit = True
 tag = False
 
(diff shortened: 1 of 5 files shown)


illusional (approved):
Gross internal methods, but I spent some time thinking about this and looking up the usages, it makes sense to use this method directly. It probably would be good to remove the leading underscore and update usages. But a problem for another day.

Can you add a bumpversion commit, then this is all good.

Thanks @MattWellie

PR #168, opened 2024-05-15
Update workflow options for cromwell gcs batch

illusional:
As far as I can tell, this is undocumented, but found by trawling around the Cromwell source: https://github.com/broadinstitute/cromwell/blob/c8d36d5c7df25f79e87ea5f6373ab5ee1556a6fb/supportedBackends/google/batch/src/main/scala/cromwell/backend/google/batch/models/GcpBatchWorkflowPaths.scala#L20

Suggested a documentation change here: https://github.com/broadinstitute/cromwell/pull/7435

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -2,7 +2,7 @@ name: Test
 on: [push]
 
 jobs:
-  lint:
+  unit-tests:
     runs-on: ubuntu-latest
     defaults:
       run:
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yaml
```diff
@@ -2,7 +2,7 @@ name: Test
 on: [push]
 
 jobs:
-  lint:
+  unit-tests:
```
illusional, line 5: Unrelated, but a little annoying when looking at GH action status

MattWellie (approved):
LGTMatt

nevoodoo (approved):
Having that chat def helped the context for this, looks good!

PR #169, opened 2024-05-17
Relaxing pinning to botocore.

milo-hyben:


illusional (approved):
Need a bump version commit here, then we should release!

milo-hyben:
When I was testing had an issue with boto3 being pinned as well, so relaxed that one too.

PR #173, opened 2024-09-14
Add sonarqube scan to test workflow

violetbrina:


.github/workflows/test.yaml
```diff
@@ -23,12 +24,24 @@ jobs:
     - name: "Run unit tests"
       id: runtests
       run: |
-        coverage run -m unittest discover -p 'test*.py' -s '.'
+        coverage run -m pytest/ --junitxml=test-execution.xml
```
nevoodoo: Maybe there's a typo here with the extra slash? Might be causing the error

nevoodoo (changes requested):

.github/workflows/test.yaml
violetbrina: Fixed!

nevoodoo (approved):

jmarshall:
I realise I am late to the party, but: What is the reason for removing codecov?

violetbrina:
Mostly that it's totally redundant. I didn't see much point in keeping it around if the coverage is in sonarqube anyway. Good question though. If you feel there is a codecov shaped whole in the repos let me know and we can always add it back. But it was only on 2 maybe 3 of our repos anyway.

PR #176, opened 2024-10-22
Quit after aborting

MattWellie:
Noted by @EddieLF  - the tasks were aborted successfully, but the watcher itself isn't terminated. This results in a cycle of attempting to abort an aborted run over and over 🙃


EddieLF (approved):
Nice simple fix 👍 ➕

PR #178, opened 2024-12-17
Batch.write output takes paths

MattWellie:
- Hail's write_output method requires the output path to be a String
- Most of our pipeline work uses Paths
- We have 23 instances in production-pipelines of having to do this string conversion inside the method call: https://github.com/search?q=repo%3Apopulationgenomics%2Fproduction-pipelines+str%28output+write_output&type=code

And there's probably a few more where we've done the String conversion before the call.

We're using the `cpg-utils.hail_batch.Batch` wrapper class as our pipeline `Batch`, so inserting this method here means we can use Paths or Strings as the output path variable without getting shouted at by Hail. This just takes the same input arguments as the true method interface, guarantees a String representation of the output path, and calls up to the `Super.write_output` method

It would be nice to change this upstream in Hail, but that would incur a long release time, and would not be mutually exclusive with this patch

### DIFF
--- a/cpg_utils/hail_batch.py
+++ b/cpg_utils/hail_batch.py
@@ -223,6 +223,7 @@ class Batch(hb.Batch):
     def write_output(self, resource: _resource.Resource, dest: str):
         """
         Write the output of a resource to a destination.
+        Includes an explicit String cast to prevent failures when calling with Cloud/PosixPath objects
         """
         return super().write_output(resource, str(dest))
 
--- a/cpg_utils/hail_batch.py
+++ b/cpg_utils/hail_batch.py
@@ -17,6 +17,7 @@ import hail as hl
 import hailtop.batch as hb
 from hail.backend.service_backend import ServiceBackend as InternalServiceBackend
 from hail.utils.java import Env
+from hailtop.batch import resource as _resource
 from hailtop.config import get_deploy_config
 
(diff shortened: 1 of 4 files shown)


Alexander-Stuckey (commented):
Looks good to me, will wait for @jmarshall  as well just in case there's something I'm missing

cpg_utils/hail_batch.py
```diff
@@ -219,6 +220,13 @@ def _process_job_attributes(
         fixed_attrs = {k: str(v) for k, v in attributes.items()}
         return name, fixed_attrs
 
+    def write_output(self, resource: _resource.Resource, dest: str):
```
jmarshall, line 223: You can just spell this as `hb.Resource`, and `dest` should be `str | Path`.

jmarshall:
This is similar to PR #166, but for `write_output()`. I'm certainly in favour of these Hail routines whose parameters are filepaths just doing the right thing for Paths as well as strs. We ended up not needing to apply #166 as the upstream version appeared in Hail 0.2.131.

I expect that the upstream hail-is/hail#14770 will appear in the next Hail release, which I expect will be fairly early in the new year. I'm ambivalent about deploying a new cpg-utils just for this as that wait is not all that long, but it could happily tag along with something else that required a cpg-utils deployment.

PR #179, opened 2025-03-19
Update from gsutil to gcloud

MattWellie:
For about 2 years the driver image has contained GCloud, but we're still using gsutil

https://cloud.google.com/blog/products/storage-data-transfer/new-gcloud-storage-cli-for-your-data-transfers

Specific to the `retry_gs_cp` method, we were disabling checksums with `-o GSUtil:check_hashes=never`, not sure why that was. We can trigger the same behaviour in gcloud with `gcloud config set storage/check_hashes never` but... checking data integrity sounds like something we want to do?

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 5.2.0
+current_version = 5.2.1
 commit = True
 tag = False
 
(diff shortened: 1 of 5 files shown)


Alexander-Stuckey (approved):
Looks all good to me

PR #180, opened 2025-04-15
Use a PEP625-compliant sdist filename and PyPI Trusted Publishing

jmarshall:
Similar workflow updates to populationgenomics/analysis-runner#717:—

Ensure the setuptools used is a version that produces a PEP625-compliant sdist filename, i.e., `cpg_utils-*` rather than `cpg-utils-*`. (This requires setuptools ≥69.3.0; the current release is 78.x.) For later Python versions, actions/setup-python does not install setuptools at all, so we will always need to have a setuptools installation step in future. Fixes SET-305.

Also use Trusted Publishing to upload to PyPI, similarly to. As per best practices, separate uploading into a separate job (which requires uploading/download artifacts to communicate between jobs) so that only the PyPI publishing step gets the credentials.

### DIFF
--- a/.github/workflows/package.yaml
+++ b/.github/workflows/package.yaml
@@ -7,16 +7,17 @@ on:
 jobs:
   package:
     runs-on: ubuntu-latest
-    environment: production
-    defaults:
-      run:
-        shell: bash -l {0}
     steps:
       - uses: actions/checkout@v4
 
       - uses: actions/setup-python@v5
         with:
           python-version: "3.10"
+
+      - name: Install prerequisites
+        run: |
+          # Ensure setuptools produces a PEP625-compliant sdist filename
+          pip install --upgrade setuptools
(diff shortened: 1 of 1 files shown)


jmarshall:
A trusted publisher flow has been set up at PyPI as follows:

populationgenomics/cpg-utils trusted publisher

and a `pypi` environment that only accepts deployments from the **main** branch should be set up in this GitHub repo.

nevoodoo:
Nice! I've setup the environment protections for `pypi` on `main`. Looking good!

nevoodoo (approved):

jmarshall:
Despite the `skip-existing: true`, I'm expecting to see a minor failure because it'll upload the same version again and not see it as skippable because of the `-` vs `_` difference in the filename. But it'll be fine in future…

PR #181, opened 2025-04-17
Add a version parameter to image_path()

jmarshall:
When given an explicit version, constructs the production image path by looking up the specified tag (specified as the `version` parameter) in the GCP artifact registry and returning the URI returned by the API (modified to use the tag rather than the raw sha256 checksum).

Currently still allows the existing behaviour, for compatibility with existing usage.

Add image lookup functionality to `cpg_utils.cloud`: note that the GCP API appears to only support scanning all images in a repository, so we cache them at the first lookup. Also increase the page_size (default 30) to reduce the overhead; e.g., on our current production images repository (~2000 images), this reduces the `_ensure_image_tags_loaded()` impact from ~1 minute to 10 seconds.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 5.2.1
+current_version = 5.3.0
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -5,7 +5,7 @@ on:
       - main
 
 env:
-  VERSION: 5.2.1
+  VERSION: 5.3.0
 
 permissions:
   contents: read
--- a/cpg_utils/cloud.py
+++ b/cpg_utils/cloud.py
@@ -5,7 +5,9 @@ import os
 import re
 import subprocess
 import traceback
-from typing import Any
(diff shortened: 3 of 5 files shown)


milo-hyben (approved):
Looking good John!
I am not a big fan of global variables, but I can see populating _repo_image_tags from ArtifactRegistry each time find_image is called would slow down AR jobs.

vivbak:
Let's make sure we merge at the same time as the migration!

jmarshall:
Turns out there **is** a way in the GCP API to query the image tags on (all images with) a particular image name: see [ListTagsRequest](https://cloud.google.com/python/docs/reference/artifactregistry/latest/google.cloud.artifactregistry_v1.types.ListTagsRequest). However this took (from my Mac, i.e., from outwith GCP) 2–3 seconds for a single [`example request`](https://artifactregistry.googleapis.com/v1/projects/cpg-common/locations/australia-southeast1/repositories/images/packages/samtools/tags?alt=json), so 10 seconds to get everything that might be used in a batch is still a win.

vivbak:
Just a note, given the backwards compatibility we probably don't need to wait for the migration to merge!

PR #182, opened 2025-06-06
Adds class modules to __all__ for discovery

MattWellie:
Change: Adds all modules in cpg_utils to the top level package `__init__.py` to aid in module discovery using simplified inputs

---

Based on various code style discussions, we may want to simplify imports and use full paths to modules/methods in imported libraries. My understanding was that instead of this (deliberately horrible) example:

```
from Pathlib import Path as pathlib_Path
from cpg_utils import Path, to_path
from cpg_utils.hail_batch import init_batch

path_obj: Path = to_path('the/actual/path')
pathlib_obj: pathlib_Path = pathlib_Path('another/path')
init_batch(**)
```

the clarified import/reference would be more like:

```
import Pathlib
import cpg_utils

path_obj: cpg_utils.Path = cpg_utils.to_path('the/actual/path')
pathlib_obj: pathlib.Path = pathlib.Path('another/path')
cpg_utils.hail_batch.init_batch(**)
```

Or.. I've completely misunderstood the intention...

If this is along the line of thinking, we have an issue in that our codebases do a bad job of populating the `__all__` attribute, leading to problems discovering module components unless they're imported in a fully qualified way. 

Screenshot 2025-06-06 at 1 12 14 PM

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 5.3.0
+current_version = 5.3.1
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -5,7 +5,7 @@ on:
       - main
 
 env:
-  VERSION: 5.3.0
+  VERSION: 5.3.1
 
 permissions:
   contents: read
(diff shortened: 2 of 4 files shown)


MattWellie:
nah

PR #183, opened 2025-06-09
Compress sequencing_groups job attribute when it is very large

jmarshall:
Production-pipelines etc workflows generator-side implementation for SET-565.

Add `_pack_attribute()` to compress attribute values if necessary. (Suppress annoying ruff warnings about "incomprehensible" numeric constants.)

Pop `'sequencing_groups'` from the incoming attributes as we may not replace the existing item later. Retain `or []` to continue to guard against the unlikely event of the raw incoming value being None rather than `[]`.

This compresses `sequencing_groups` (and is flexible enough for other attributes in future if necessary) for transport through the Hail database to the billing aggregator. populationgenomics/cpg-infrastructure#308 processes any `sequencing_groups_gzip` attributes it sees, and should be merged and deployed before this PR.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 5.3.0
+current_version = 5.4.0
 commit = True
 tag = False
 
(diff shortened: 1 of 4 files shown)


milo-hyben (approved):
Looking good to me!

PR #185, opened 2025-07-17
Tag URIs are colon-delimited, unlike at-delimited digest URIs

jmarshall:
Fix `tag_uri`, which will be returned by `image_path(pkg, version)` and should be delimited with `:` rather than the `@` used to delimit the digest URI it is reconstructed from:

Screenshot 2025-07-17 at 19 00 48

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 5.4.0
+current_version = 5.4.1
 commit = True
 tag = False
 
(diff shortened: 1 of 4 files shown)


jmarshall:
See e.g. [batch 628755](https://batch.hail.populationgenomics.org.au/batches/628775/jobs/1) which shows the current syntax is invalid. Oops…

dancoates:
Do we need to support specifying a sha? In that case the format would be 

```python
f'{base_uri}@sha256:{tag}',
```

dancoates:
Oh wait I see, the tags come from the registry, not from the user. This should be all good then

dancoates (approved):

PR #186, opened 2025-08-13
Avoid modifying incoming attributes when compressing 'sequencing_groups'

jmarshall:
Multiple jobs in the batch might reference the same attributes dict object. Avoid modifying the dict object (e.g. with pop() or update()) to avoid changing the attributes of subsequently processed jobs.

Add a comment to that effect so the next person to modify this function is aware of this part of its contract with its callers!

Compare [batch 631425](https://batch.hail.populationgenomics.org.au/batches/631425/jobs/1) prior to this fix with [batch 631427](https://batch.hail.populationgenomics.org.au/batches/631427/jobs/1) which includes it. (Both use dev cpg-flow images built from populationgenomics/cpg-flow#123. See in particular the `CPGUTILS _process_job_attributes pre  Filter Evens` lines in the logs.)

### DIFF
--- a/cpg_utils/hail_batch.py
+++ b/cpg_utils/hail_batch.py
@@ -189,12 +189,16 @@ class Batch(hb.Batch):
 
         self.total_job_num += 1
 
+        # Multiple jobs in the batch might reference the same attributes dict
+        # object. Avoid modifying the dict object (e.g. with pop() or update())
+        # to avoid changing the attributes of subsequently processed jobs.
+
         attributes = attributes or {}
         stage = attributes.get('stage')
         dataset = attributes.get('dataset')
         sequencing_group = attributes.get('sequencing_group')
         participant_id = attributes.get('participant_id')
-        sequencing_groups: set[str] = set(attributes.pop('sequencing_groups', []) or [])
+        sequencing_groups: set[str] = set(attributes.get('sequencing_groups') or [])
         if sequencing_group:
             sequencing_groups.add(sequencing_group)
         part = attributes.get('part')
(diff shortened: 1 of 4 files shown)


cpg_utils/hail_batch.py
```diff
@@ -189,12 +189,16 @@ def _process_job_attributes(
...
+        # object. Avoid modifying the dict object (e.g. with pop() or update())
+        # to avoid changing the attributes of subsequently processed jobs.
+
         attributes = attributes or {}
         stage = attributes.get('stage')
         dataset = attributes.get('dataset')
         sequencing_group = attributes.get('sequencing_group')
         participant_id = attributes.get('participant_id')
-        sequencing_groups: set[str] = set(attributes.pop('sequencing_groups', []) or [])
+        sequencing_groups: set[str] = set(attributes.get('sequencing_groups') or [])
```
dancoates, line 201: Should this be `attributes.get('sequencing_groups', [])`? With this update it will raise an exception if the key `sequencing_groups` doesn't exist, whereas previously it would not.

dancoates (commented):
Looks good, just one small note on a potential bug

cpg_utils/hail_batch.py
jmarshall: This reverts the line back to what it was prior to PR #183. If that key doesn't exist, it'll be `set(None or [])` which will be an empty set as expected.

Personally I'd rather write it as `… = set(attributes.get('sequencing_groups', []))` but that will newly raise an exception if someone somehow passes in `{'sequencing_groups': None}` whereas the `or []` version won't.

In summary, 🤮 🤷 …

dancoates: > If that key doesn't exist, it'll be set(None or [])

I don't think that's right, if the key doesn't exist it will raise a KeyError. To allow for both the key not existing, and it being `None` you could do `set(attributes.get('sequencing_groups', []) or [])`.

But as you say, this is reverting back to what it was previously so presumably what you've got at the moment will be ok unless there's been other updates in the meantime that would cause `sequencing_groups` to not exist.

dancoates: Wait, I'm wrong - pop raises a KeyError but get doesn't. Ignore me

dancoates (approved):

PR #187, opened 2025-12-04
Use newer Slack SDK function instead of obsolete files_upload

jmarshall:
Slack's `files.upload` endpoint has been [deprecated for eighteen months](https://docs.slack.dev/changelog/2024-04-a-better-way-to-upload-files-is-here-to-stay/) and as of 12 November 2025 is no longer functional. Fortunately the SDK provides a `files_upload_v2()` wrapper around the replacement endpoints that is used very similarly.

The `files_upload_v2()` function is underdocumented in the reference documentation, but is reasonably documented in the [SDK source code](https://github.com/slackapi/python-slack-sdk/blob/4601f8a82836cf649ffe0f5d7b08061882d6361f/slack_sdk/web/client.py#L3966). This revised code has been tested against a test slack instance (`upload_file('File text.\n', 'See attachment')`) and posted as expected:

via files_upload_v2()

Context: https://centrepopgen.slack.com/archives/C03FZL2EF24/p1764715613227479
Fixes [SET-861](https://cpg-populationanalysis.atlassian.net/browse/SET-861).

We will need to rebuild the analysis-runner driver image with this cpg-utils update to get this fix propagated to the storage visualisation (which uses `upload_file()`).

[SET-861]: https://cpg-populationanalysis.atlassian.net/browse/SET-861?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 5.4.2
+current_version = 5.4.3
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -5,7 +5,7 @@ on:
       - main
 
 env:
-  VERSION: 5.4.2
+  VERSION: 5.4.3
 
 permissions:
   contents: read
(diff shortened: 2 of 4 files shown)


jmarshall:
While we're here, add some parameters to allow customising the “Uploaded file” caption, etc.

dancoates (approved):
Looks good!

PR #188, opened 2025-12-22
Expose parsed docker images

MattWellie:
Adds a public method to sit on top of the existing artifactory parsing code, allowing a user to view the available tags for a project-location-repository-image combination. The `_repo_image_tags` remains a private variable, but we now have a 'public' method to expose all tags for a known image.

Additional: alteration to types in DockerImage 

- image size should be an int
- build_time is natively a [DateTimeWithNanoseconds](https://googleapis.dev/python/google-api-core/latest/helpers.html#google.api_core.datetime_helpers.DatetimeWithNanoseconds), but I've swapped it to a standard Python datetime object with an attached timezone

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 5.4.3
+current_version = 5.4.4
 commit = True
 tag = False
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -5,7 +5,7 @@ on:
       - main
 
 env:
-  VERSION: 5.4.3
+  VERSION: 5.4.4
 
 permissions:
   contents: read
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,16 +24,16 @@ repos:
     hooks:
       - id: cpg-id-checker
 
-  - repo: https://github.com/ambv/black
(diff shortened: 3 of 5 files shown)


EddieLF (approved):
LGTM

jmarshall:
(We don't have a _CODEOWNERS_ file in this repository yet, but to be clear: @populationgenomics/software-team will also review this before it should be merged.)

jmarshall:
BTW do you have a draft PR somewhere showing how you want to use this API?

MattWellie:
> BTW do you have a draft PR somewhere showing how you want to use this API?

Not exactly, we have a quirky situation where `mitoreport` is installed in an image, but also requires an internal database update to remain up to date. We are planning for a fun situation where:

- image is periodically rebuilt, e.g. monthly
- the software version is not updated, but the tag is
- workflows using `mitoreport` would want to search for the latest available increment of a known version of the mitoreport image with each run

So assuming we're using version `1.2.3`, the workflow would use this endpoint to search for all tags with a version starting `1.2.3-*`, then choose either the highest increment or the latest datetime for the workflow.

---

Outside of that specific workflow context it feels like a good move to enable access to the currently private objects that have done all the heavy lifting of contacting the images API. The dictionary is populated but contains no public access points, so we can't even check if a given tag exists using this interface, it all requires foreknowledge of tags.

PR #190, opened 2026-03-15
Add desktop oauth flow to `get_google_identity_token` method

dancoates:
This method previously would get a user's Application Default Credentials in the case where they had a locally authenticated account. But ADC credentials can't be used to make a request to an IAP secured resource. That requires a OAuth 2.0 ID token which can be obtained by going through the Desktop OAuth flow.

This requires having the user visit a page in a browser which then redirects back to a local URL with an authorization code. The ID Token can then be fetched from Google with a simple request which returns the ID Token and refresh token. These tokens are then cached in a local config dir which allows for subsequent requests to avoid the need to perform the full OAuth flow again.

This flow is only needed for human users, service accounts and compute instances can still retrieve an ID token without the requirement of an interactive flow. This new flow is also implemented conditionally to ensure backwards compatibility with other usages of this method which expect a request from a user to return ADC credentials rather than an ID token.

### DIFF
--- a/cpg_utils/cloud.py
+++ b/cpg_utils/cloud.py
@@ -1,18 +1,20 @@
 """Convenience functions related to cloud infrastructure."""
 
+import contextlib
 import json
 import os
 import re
 import subprocess
+import time
 import traceback
 import urllib.parse
 from collections import defaultdict
+from pathlib import Path
 from typing import Any, NamedTuple
 
 # pylint: disable=no-name-in-module
 import google.api_core.exceptions
 import google.auth.transport
-import google.oauth2
 from deprecated import deprecated
 from google.auth import (
     credentials as google_auth_credentials,
@@ -31,6 +33,7 @@ from google.auth.transport import requests
 from google.cloud import artifactregistry, secretmanager
 from google.oauth2 import credentials as oauth2_credentials
 from google.oauth2 import service_account
+from google_auth_oauthlib.flow import InstalledAppFlow
 
 _CLOUD_SDK_MISSING_CREDENTIALS = """\
 Your default credentials were not found. To set up Application Default Credentials, \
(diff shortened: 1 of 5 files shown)


hrrsheen (approved):
Thanks for adding those tests. LGTM.

cpg_utils/cloud.py
```diff
@@ -175,9 +178,211 @@ def find_image(repository: str | None, name: str, version: str) -> DockerImage:
...
+            with open(self.TOKEN_CACHE_FILE) as f:
+                full_cache = json.load(f)
+
+            # The cache is a nested dict: {iap_client_id: {desktop_client_id: { ...tokens... }}}
+            iap_cache = full_cache.get(self.iap_client_id, {})
+            return iap_cache.get(self.desktop_client_id, {})
+        except (json.JSONDecodeError, KeyError):
+            return {}
+
+    def _save_cache(
```
ckrishk, line 215: Nit : very likely could fail during concurrent requests but don't think we will run into those usecases often.

cpg_utils/cloud.py
```diff
@@ -175,9 +178,211 @@ def find_image(repository: str | None, name: str, version: str) -> DockerImage:
...
+
+        if refresh_token:
+            try:
+                id_token = self._get_id_token_from_refresh_token(request, refresh_token)
+                self._cache_id_token(id_token)
+                self.token = id_token
+                return
+            except RuntimeError as e:
+                # If the refresh token is expired or revoked, clear it and fall through to do the flow again.
+                if 'expired' in str(e).lower() or 'revoked' in str(e).lower():
```
ckrishk, line 366: Not too fussed, but may be worth checking for error code instead of string status ? I am assuming we want to reject all errors other than this.

ckrishk (commented):
Looks good to me except for a couple of nits

PR #194, opened 2026-06-26
Adds analysis registration script

MattWellie:
A candidate collection of methods/standalone script for registering outputs in metamist. This general process has been trialled in NF/Seqera workflows, but required the script to leave inside each workflow. Migrating this code upstream will mean each CPG-Utils install makes this available.

This is not a script per se, it's a parameterised metamist interaction for registering analyses. The previous incarnation in the seqera-trial repository was a script, because that's what NextFlow required. This implementation takes that same interaction, keeps a script-like entrypoint, and also adds the same functionality as an importable method. The exact same logic is exposed both ways, making this solution usable for both code-accessed and cli-accessed registrations.

- this started with @dancoates' OG standalone metamist script [here](https://github.com/populationgenomics/cpg-seqera-trial-workflows/blob/main/workflows/metamist_demo/bin/update_metamist.py)
- that was boosted [here](https://github.com/populationgenomics/cpg-seqera-trial-workflows/blob/cohort_workflow/workflows/cohort_demo/bin/update_metamist.py), adding flexibility to register any analysis type, populating analysis meta, and adding secondary files to analysis records, which didn't previously exist in CPG-Flow workflows.
 
This file contains a `__main__` entrypoint, or a direct-use `create_new` method. If called as a script the CLI looks like:

```
python -m cpg_utils.metamist_registration \
    --project <project_name> \
    --output "<path to a primary output file>" \
     --type <analysis-type> \
     --cohorts COHABC COHDEF *or* --sgs CPGABC CPGDEF \
     [--meta stage=test sequencing_type=genome] \
     [--secondary index=<path to an index file>] 
```

If run via import in code, usage looks like:
```
from cpg_utils.metamist_registration import create_new

create_new(
    project=<project>,
    output=<path/to/primary/output>,
    analysis_type=cram,
    meta={'stage': 'name', 'sequencing_type': 'genome'},
    secondary={'index': <path to an index file>}
    cohorts=['COHABC', 'COHDEF'],
)
```

This interface takes:
- either SG or Cohorts (lists of one or more, but each analysis can only be aligned to SG _or_ Cohorts)
- a main analysis output path (string, typically a GS-path)
- an unlimited number of secondary files, each with a corresponding type (the type must already exist in metamist)
- an unlimited number of additional metadata key=values to be appended to the analysis record's meta (meta is not populated for any secondary files)

The result is:
- an analysis record of the indicated **type**, inside the **project**
- the analysis.meta dictionary is populated with all key-value pairs 
- if required, the analysis.outputs contains a nested secondary_files dict, each with a key identifying the secondary output type. 

The analysis record 388733 has been generated by this process as a demonstration, go check it out.

> NB by writing analysis records in a way which enables secondary files, the `analysis.output` field is always an empty String. This is a deprecated field, so we're probably comfortable with this, but some processes/workflows still prioritise reading the output String instead of _`analysis.outputs.path`_

### DIFF
--- a/cpg_utils/metamist_registration.py
+++ b/cpg_utils/metamist_registration.py
@@ -50,6 +50,7 @@ import argparse  # noqa: I001
 import json
 import sys
 
+from cpg_utils.existence_checks import exists
 from metamist import exceptions, graphql
 
 
@@ -89,6 +90,17 @@ def parse_cli_kv(input_kv: list[str]) -> dict[str, str]:
             raise ValueError(f'Duplicate key provided: {key}')
         broken_kv[key] = value
     return broken_kv
+
+
+def find_missing_files(primary: str, secondary: dict[str, str]) -> set[str]:
+    """For the primary and secondary files, detect if any are missing. Return all missing Paths."""
+    missing_files: set[str] = set()
+    if not exists(primary):
+        missing_files.add(primary)
+    for filepath in secondary.values():
+        if not exists(filepath):
+            missing_files.add(filepath)
+    return missing_files
 
 
 def create_output_block(
@@ -199,10 +211,20 @@ def create_new(
     """
     # fail if the cohorts and sgs are both applied, or if neither is applied
     if cohorts and sgs:
-        raise Exception('Cannot specify both --cohorts and --sgs')
+        raise ValueError(
+            'Cannot specify both --cohorts and --sgs CLI parameters for a single Analysis.',
+        )
 
     if not (cohorts or sgs):
-        raise Exception('Must specify either --cohorts or --sgs')
+        raise ValueError(
+            'You must specify either --cohorts or --sgs for a single Analysis object.',
+        )
+
+    if missing_files := find_missing_files(primary=output, secondary=secondary or {}):
+        missing_file_string = ', '.join(sorted(missing_files))
+        raise ValueError(
+            f'Missing files detected: {missing_file_string}.\nThis can only be used for extant files.',
+        )
 
     outputs = create_output_block(
         primary=output,
--- a/cpg_utils/metamist_registration.py
+++ b/cpg_utils/metamist_registration.py
@@ -54,6 +54,7 @@ from cpg_utils.existence_checks import exists
 from metamist import exceptions, graphql
 
 
+GS_PREFIX = 'gs://'
 UPDATE_QUERY = """
     mutation updateAnalysis($project: String!, $analysis:AnalysisInput!) {
         analysis {
@@ -95,12 +96,19 @@ def parse_cli_kv(input_kv: list[str]) -> dict[str, str]:
 def find_missing_files(primary: str, secondary: dict[str, str]) -> set[str]:
     """For the primary and secondary files, detect if any are missing. Return all missing Paths."""
     missing_files: set[str] = set()
-    if not exists(primary):
-        missing_files.add(primary)
-    for filepath in secondary.values():
+    for filepath in [primary, *list(secondary.values())]:
         if not exists(filepath):
             missing_files.add(filepath)
     return missing_files
+
+
+def find_non_gcs_files(primary: str, secondary: dict[str, str]) -> set[str]:
+    """Checks for a `gs://` prefix on all files."""
+    non_gs_files: set[str] = set()
+    for filepath in [primary, *list(secondary.values())]:
+        if not filepath.startswith('gs://'):
+            non_gs_files.add(filepath)
+    return non_gs_files
 
 
 def create_output_block(
@@ -224,6 +232,11 @@ def create_new(
         missing_file_string = ', '.join(sorted(missing_files))
         raise ValueError(
             f'Missing files detected: {missing_file_string}.\nThis can only be used for extant files.',
+        )
+    if non_gcs_files := find_non_gcs_files(primary=output, secondary=secondary or {}):
+        missing_file_string = ', '.join(sorted(non_gcs_files))
+        raise ValueError(
+            f'Non-GCS files detected: {missing_file_string}.\nThis can only be used for files stored in GCS.',
         )
 
     outputs = create_output_block(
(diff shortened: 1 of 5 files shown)


cpg_utils/metamist_registration.py
```diff
@@ -0,0 +1,238 @@
...
+        print('DRY RUN, would have POSTed the following analysis data:')
+        print(json.dumps(variables, indent=2))
+        sys.exit(0)
+
+    try:
+        result = graphql.query(UPDATE_QUERY, variables)
+        print(f'Successfully recorded analysis to Metamist: {result}')
+    except exceptions.ApiException as e:
+        print(f'Error executing GraphQL query: {e}', file=sys.stderr)
+        sys.exit(1)
```
Johnnyassaf: These sys exit commands might break other scripts if registration isn't the last thing a script does

jmarshall:
At the moment cpg-utils does not depend on metamist; meanwhile metamist does use cpg-utils. Also cpg-utils is purely a library for use from Python code and does not currently contain any scripts.

So this doesn't seem like a great fit for cpg-utils. Are there other places this facility could be added instead  to make it widely available?

MattWellie:
Strategically this is both a script and an importable process - allowing its use as a standalone script doesn't affect anything else in the library - taking behaviour we only previously accessed via prod-pipes and CPG-Flow, and casting it as a generically accessible function with a simple interface.

This core function is required across all workflows during the migration away from CPG-Flow/hail batch as our runner, so it should live somewhere communal, and CPG-utils is already ubiquitous either directly or as a dependency in all our workflows.

cpg_utils/metamist_registration.py
MattWellie: Agreed. I've removed the `exit(0)`, and made the `exit(1)` a re-raise of the ApiException. If it's called as a script it would hard fail, if it's called through code it could be handled/caught.

Johnnyassaf (commented):
From a design perspective I understand not wanting to introduce a dependency on metamist, and not wanting to risk a cyclic dependency.  

We don't really have a communal scripts repo though and I think this script fits in cpg-utils as CPG communal code that can live as a script (just register jobs when you're done with work) or as a function (register outputs as part of a job).  As long as we ensure that the metamist functions imported here don't import this cpg-utils script in the future, it should be fine.

jmarshall:
I guess the onus is on the PR's proponents to demonstrate that the cyclic dependency is not an issue.

I see this as yet more grist for the mill of sorting out a suitable location for scripts.

MattWellie:
I'm not sure where this talk of cyclic dependencies is coming from. There's no install-time cyclic issues, pip or any other package manager treats this as an "other and self" dependency resolution, not a recursion risk. There are cpg-utils imports within the metamist library, but a new import of `cpg_utils.metamist_registration` would have to be added for that to be a risk. Specifically it would have to be added in the graphql or exceptions modules, which would be wildly inappropriate. 

This is adding new functionality to be used globally across the CPG, which is the exact remit of cpg-utils. It is not a `script` like we have previously discussed for a scripts repository (i.e. a one-shot isolated process), it is a behaviour which we need in all of our separate workflows. The ability to operate it via import or as a standalone entrypoint is a benefit (cpg-flow sets up workflows in python, nextflow stages are a series of command line calls), but this should not be confused for `a script`

Johnnyassaf (approved):
LGTme

cpg_utils/metamist_registration.py
```diff
@@ -0,0 +1,237 @@
...
+        action='store_true',
+        help='Dry run, print only.',
+    )
+    args = parser.parse_args()
+
+    # do some transformation of the input meta/secondary file dictionaries
+    meta_dict = parse_cli_kv(args.meta)
+    secondary_dict = parse_cli_kv(args.secondary)
+
+    # call the main analysis generation method
```
Alexander-Stuckey, line 184: Suggest that both outputs and any file passed as a secondary file be checked to ensure that they are cloudpaths and also exist in the cloud (`gs://` string and `path.exists()` or so)

cpg_utils/metamist_registration.py
```diff
@@ -0,0 +1,237 @@
...
+        analysis_type: str, the analysis type (must be from valid enum in metamist)
+        meta: optional dict, if provided this will be added as the analysis.meta dictionary
+        secondary: optional dict, if provided each element of this {type: string} dict will be added as a secondary file
+        cohorts: optional list, COHort IDs to attribute the analysis to. Mutually exclusive with sgs
+        sgs: optional list, Sequencing Group IDs to attribute the analysis to. Mutually exclusive with cohorts
+        dry: bool, if True, payload is printed, but not sent
+    """
+    # fail if the cohorts and sgs are both applied, or if neither is applied
+    if cohorts and sgs:
+        raise Exception('Cannot specify both --cohorts and --sgs')
```
Alexander-Stuckey: Perhaps `ValueError` instead of bare `Exception`

cpg_utils/metamist_registration.py
```diff
@@ -0,0 +1,237 @@
...
+
+    if not (cohorts or sgs):
+        raise Exception('Must specify either --cohorts or --sgs')
```
Alexander-Stuckey: Likewise

Alexander-Stuckey (changes requested):
Few comments, mostly looks good

cpg_utils/metamist_registration.py
MattWellie: yup, good idea. This came up in conversation earlier as cpg-flow does do an existence check, and registering a file which doesn't exist fails to trigger the first-class-files logic

MattWellie: sliiightly expanded the remit here - I've pulled in the directory-level existence checking from CPG-Flow. Mainly because it's cool functionality and I'd want it accessible in a post-cpg-flow world, but also to use here where I suspect most secondary files will be in the same directory (e.g. data file and index), so the single call + check against cache will be useful

Alexander-Stuckey (approved):
Looks good to me

